PR #20, opened 2020-06-15
Update pulsar image to 2.6.0

sijie:


sijie:
the image is the current release tag. After the release is officially published, we need to change to the 2.6.0 release.

PR #22, opened 2020-06-22
Allow to change broker service account annotations

jujugrrr:
### Motivation

We need to be able to change annotation to inject AWS IAM role (EKS based deployment).
https://docs.aws.amazon.com/eks/latest/userguide/specify-service-account-role.html

With 2.6.0 and this annotation change we were able to use Tiered Storage with S3 and EKS/IAM(OIDC).

e.g : 
```
  annotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::66666:role/my-iam-role-with-s3-access
```
values.yaml
```
broker:
  service_account:
    annotations:
      eks.amazonaws.com/role-arn: arn:aws:iam::66666:role/my-iam-role-with-s3-access
```
### Modifications

Added a value to allow to change annotations fro broker service account.
I've tried following style from other part of the code.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/broker-service-account.yaml
+++ b/charts/pulsar/templates/broker-service-account.yaml
@@ -26,4 +26,8 @@ metadata:
   labels:
     {{- include "pulsar.standardLabels" . | nindent 4 }}
     component: {{ .Values.broker.component }}
+  annotations:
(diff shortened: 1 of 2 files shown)


jujugrrr:
Thanks @sijie

PR #24, opened 2020-06-23
Remove double quotes from the environment variables

sijie:
*Motivation*

Some of the environment variables still use double quotes. They result in the following

```bash
Could not find or load main class "
```

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -496,9 +496,7 @@ autorecovery:
   ##
   configData:
     BOOKIE_MEM: >
-      "
       -Xms64m -Xmx64m
-      "
 
 ## Pulsar Zookeeper metadata. The metadata will be deployed as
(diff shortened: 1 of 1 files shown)


jujugrrr:
@sijie I've noticed /pulsar/conf/pulsar_env.sh is returning errors without the quote(broker logs):

```
broker xxx-pulsar-broker-0.tf-res-sandbox-pulsar-broker.xxx-pulsar.svc.cluster.local znode still exists ... check in 10 seconds ...
/pulsar/conf/pulsar_env.sh: line 45: -Xmx256m: command not found
```

I'm not sure if it's actually creating errors somewhere else, but is it expected?

sijie:
@jujugrrr I think the helm chart still tries to substitute the environment variables for some init containers in the environment file. This results in the error messages but it doesn't impact the running containers. Let me submit a fix.

jujugrrr:
Thanks a lot!

PR #28, opened 2020-06-26
Don't substitute environment variables

sijie:
*Motivation*

environment variables are already taken by bash scripts. We don't need to substitute them.


rvashishth:
@sijie would it be worth to keep updating the [chart version](https://github.com/apache/pulsar-helm-chart/blob/master/charts/pulsar/Chart.yaml#L24) as mentioned in [helm docs](https://helm.sh/docs/topics/charts/#charts-and-versioning), with each PR merge.

sijie:
@rvashishth #21 is for releasing charts. we bump the chart version when releasing a version.

PR #37, opened 2020-07-02
add targetport for grafana and manager service

rvashishth:
### Motivation

A [Kubernetes service](https://kubernetes.io/docs/concepts/services-networking/service/) can map any incoming port to a targetPort. By default and for convenience, the targetPort is set to the same value as the port field.

This change will give the flexibility to change port without customizing the chart templates.

### Modifications

- move port under service in values file
- add targetPort for grafana and pulsar_manager service

### Verifying this change

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -803,7 +803,6 @@ grafana:
   annotations: {}
   tolerations: []
   gracePeriod: 30
-  port: 3000
   resources:
     requests:
       memory: 250Mi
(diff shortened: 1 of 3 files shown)


charts/pulsar/values.yaml
```diff
@@ -813,6 +812,8 @@ grafana:
   ##
   service:
     type: LoadBalancer
+    port: 3000
```
rvashishth, line 815: as the port is for service mapping, it would be more suitable to keep it under service section.

PR #42, opened 2020-07-17
Add Ingress to Pulsar Proxy and Pulsar Manager

NiklasWagner:
### Motivation

We're hosting our Pulsar Cluster in kubernetes but most of the actual clients are running on a different infrastructure. To expose the pulsar proxy and the pulsar manager we need a ingress resource.

### Modifications

I've copied the dashboard-ingress.yaml and adjusted the variable references and added some logic to support TLS configurations.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/proxy-ingress.yaml
+++ b/charts/pulsar/templates/proxy-ingress.yaml
@@ -1,0 +1,57 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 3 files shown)


rvashishth:
@NiklasWagner would it be a good idea to create a single ingress resource for all HTTP services( proxy admin API, Grafana, and pulsar-manager) instead of creating multiple ingresses for each component.

NiklasWagner:
I'm not so sure. It would increase the complexity if there is a edge case (for example tls on a different port) and actually there are already 10 *-service.yaml files. So it doesn't hurt if there are some more ingress files :)

rvashishth:
I assume the idea behind ingress to have a single entry point in the k8s cluster. having one ingress-resource per service my not be feasible.  I was hoping something as below 

Screenshot 2020-07-17 at 6 03 51 PM

NiklasWagner:
Ah that what you mean. We're actually using a ingress controller for that which is what most people are doing. We've many ingress resources from many different charts and our nginx ingress is taking care of that. We're not creating a LB for each ingress on our cloud provider.

Additionally: Combining this in a single ingress resource would make it more complex for the ingress controller to correctly route the traffic. I think it would even make it impossible for us to work correctly since we're also using [external-dns](https://github.com/kubernetes-sigs/external-dns) to sync the hosts from the different ingress resources.

NiklasWagner:
@sijie Is there anything left to do to get this merged? We're already using these changes for a couple weeks in production and would love to switch back to this helm chart repository.

NiklasWagner:
Thanks! :+1:

sijie:
@NiklasWagner thank you for your contributions!

PR #48, opened 2020-07-30
Allow Grafana to work with a reverse proxy

toneill818:
### Motivation

Allow Grafana to be served from a sub path.  

### Modifications

- Added a config map to add extra environment variables to the grafana deployment. As the grafana image adds new features that require environment variables, this can be used to set them.
- Bumped the grafana image to allow a reverse proxy
- removed ingress annotations as they are specific to nginx, and to match all the other ingresses
- bumped the chart version as per the README 

Example values:
```
grafana:
  configData:
    GRAFANA_ROOT_URL: /pulsar/grafana
    GRAFANA_SERVE_FROM_SUB_PATH: "true"
  ingress:
      enabled: true
      port: 3000
      path: "/pulsar/grafana/?(.*)"
      annotations:
        nginx.ingress.kubernetes.io/rewrite-target: /$1
```

### DIFF
--- a/charts/pulsar/Chart.yaml
+++ b/charts/pulsar/Chart.yaml
@@ -21,7 +21,7 @@ apiVersion: v1
 appVersion: "2.6.0"
 description: Apache Pulsar Helm chart for Kubernetes
 name: pulsar
-version: 2.6.1
+version: 2.6.0-1
 home: https://pulsar.apache.org
 sources:
 - https://github.com/apache/pulsar
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -151,7 +151,7 @@ images:
     pullPolicy: IfNotPresent
   grafana:
     repository: streamnative/apache-pulsar-grafana-dashboard-k8s
-    tag: 0.0.9
+    tag: 0.0.10
     pullPolicy: IfNotPresent
   pulsar_manager:
     repository: apachepulsar/pulsar-manager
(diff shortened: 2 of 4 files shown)


charts/pulsar/Chart.yaml
```diff
@@ -21,7 +21,7 @@ apiVersion: v1
 appVersion: "2.6.0"
 description: Apache Pulsar Helm chart for Kubernetes
 name: pulsar
-version: 2.6.0
+version: 2.6.1
```
sijie, line 24: Since the chart is for pulsar 2.6.0 and we don't release pulsar 2.6.1 yet, I would suggest not bumping the version here. We can use a naming convention like `<pulsar-version>-<minor-version>` for versioning the helm chart. How does that sound?

charts/pulsar/values.yaml
```diff
@@ -856,16 +856,16 @@ grafana:
...
+  ## Grafana configMap
+  ## templates/grafana-configmap.yaml
+  ##
+  configData: {}
   ## Grafana ingress
   ## templates/grafana-ingress.yaml
   ##
   ingress:
     enabled: false
-    annotations:
```
sijie, line 864: why did you remove the default values?

toneill818: They get merged with other values if a user provides annotations. This also matches the other ingress values. So if I am not using nginx ingress, it still attaches the nginx ingress annotations.

charts/pulsar/Chart.yaml
toneill818: I can do that, the appVersion will match the pulsar version the chart is running. The helm version and pulsar version are different. Do you want them the same so its easy to know what version you are running if you use helm?

Skaronator: @sijie Thats what the "appVersion" Value in the Chart.yaml is for. The "version" key is just the version of the helm chart and is completly separate from the actual application version. And when you follow semantic versioning on your helm chart as well then these version will seperate really quickly.

Just a quick look at the Prometheus Helm Chat: Prometheus itself is at 2.20.1 right now but the helm chart is already at 11.x https://github.com/helm/charts/blob/master/stable/prometheus/Chart.yaml#L3-L4

It would be cool to have more frequent updates of this helm chart and not just wait for pulsar updates to release at the same time.

sijie: I understand the difference between appVersion and version. My recommendation is to add another version. e.g. `2.6.1-0` this means it is a helm chart released for pulsar `2.6.1`.

sijie: Yes we don't need to align the helm chart release with pulsar release.

Skaronator: Not a huge fan of that if when we want to follow the semantic versioning for this helm chart but still better than nothing. Thanks for the new release ;)

PR #54, opened 2020-08-13
Ingress optional hostname

### LINKED ISSUE #50: Ingress hostname should not be required
**Describe the bug**
When creating ingresses for services such as the proxy or pulsar-manager, the hostname value should optional. 

**To Reproduce**
1. Create an ingress without a hostname specified
```yaml
proxy:
  ingress:
    enabled: true
    path: "/pulsar-manager/"
```
2. Try installing or upgrading the chart

**Expected behavior**
An ingress should be created.

**Additional context**
I can submit a PR to make the hostname optional.

toneill818:
Fixes #50 

### Motivation
The host option is not required to setup an ingress, so I made it an optional value
### Modifications

Made setting the host optional.

### Verifying this change

- [X] Make sure that the change passes the CI checks.

### DIFF
--- a/charts/pulsar/Chart.yaml
+++ b/charts/pulsar/Chart.yaml
@@ -21,7 +21,7 @@ apiVersion: v1
 appVersion: "2.6.0"
 description: Apache Pulsar Helm chart for Kubernetes
 name: pulsar
-version: 2.6.0-3
+version: 2.6.0-4
 home: https://pulsar.apache.org
 sources:
 - https://github.com/apache/pulsar
--- a/charts/pulsar/templates/grafana-ingress.yaml
+++ b/charts/pulsar/templates/grafana-ingress.yaml
@@ -39,12 +39,15 @@ spec:
 {{ toYaml .Values.grafana.ingress.tls | indent 4 }}
 {{- end }}
   rules:
-    - host: {{ required "Grafana ingress hostname not provided" .Values.grafana.ingress.hostname }}
-      http:
+    - http:
         paths:
         - path: {{ .Values.grafana.ingress.path }}
           backend:
             serviceName: "{{ template "pulsar.fullname" . }}-{{ .Values.grafana.component }}"
             servicePort: {{ .Values.grafana.ingress.port }}
+      {{- if .Values.grafana.ingress.hostname }}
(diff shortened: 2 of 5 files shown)


Skaronator:
I think it would be much better to use the offical helm approach for defining ingress. The host is basically a array which can be empty (eg. what you want).

You can do `helm create test` and then look at the created ingress at test/template/ingress.yaml. It looks like that:

```
{{- if .Values.ingress.enabled -}}
{{- $fullName := include "test.fullname" . -}}
{{- $svcPort := .Values.service.port -}}
{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
apiVersion: networking.k8s.io/v1beta1
{{- else -}}
apiVersion: extensions/v1beta1
{{- end }}
kind: Ingress
metadata:
  name: {{ $fullName }}
  labels:
    {{- include "test.labels" . | nindent 4 }}
  {{- with .Values.ingress.annotations }}
  annotations:
    {{- toYaml . | nindent 4 }}
  {{- end }}
spec:
  {{- if .Values.ingress.tls }}
  tls:
    {{- range .Values.ingress.tls }}
    - hosts:
        {{- range .hosts }}
        - {{ . | quote }}
        {{- end }}
      secretName: {{ .secretName }}
    {{- end }}
  {{- end }}
  rules:
    {{- range .Values.ingress.hosts }}
    - host: {{ .host | quote }}
      http:
        paths:
          {{- range .paths }}
          - path: {{ . }}
            backend:
              serviceName: {{ $fullName }}
              servicePort: {{ $svcPort }}
          {{- end }}
    {{- end }}
  {{- end }}
```
And the values.yaml looks like that:

```
ingress:
  enabled: false
  annotations: {}
    # kubernetes.io/ingress.class: nginx
    # kubernetes.io/tls-acme: "true"
  hosts:
    - host: chart-example.local
      paths: []
  tls: []
  #  - secretName: chart-example-tls
  #    hosts:
  #      - chart-example.local
```

toneill818:
@Skaronator That solution still requires a host to be defined as it loops over the hosts list. The goal of this PR was to make the host an optional input. The solution you proposed would not allow for only paths to be defined for an ingress.

PR #59, opened 2020-08-27
Pass OS signals to Pulsar container process to enable graceful shutdown

lhotari:
### Changes 

- using "exec" to run a command replaces the shell process with the executed process
- this is required so that the process running in the container is able to receive OS signals
  - explained in https://docs.docker.com/develop/develop-images/dockerfile_best-practices/
    and https://docs.docker.com/engine/reference/builder/#entrypoint
- receiving SIGTERM signal is required for graceful shutdown. This is explained in https://pracucci.com/graceful-shutdown-of-kubernetes-pods.html 

This change might fix issues such as https://github.com/apache/pulsar/issues/6603 . One expectation of this fix is that graceful shutdown would allow Pulsar components such as a bookies to deregistered from Zookeeper properly before shutdown. 

### Motivation

Dockerfile best practices mention that "exec" should be used so that the process running in a container can receive OS signals. This is explained in https://docs.docker.com/develop/develop-images/dockerfile_best-practices/
    and https://docs.docker.com/engine/reference/builder/#entrypoint .  Kubernetes documention explains pod termination in https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination : "Typically, the container runtime sends a TERM signal to the main process in each container. Once the grace period has expired, the KILL signal is sent to any remaining processes, and the Pod is then deleted from the API Server ."
Currently some issues while running Pulsar are caused by the lack of graceful shutdown. Graceful shutdown isn't happening at all since the Pulsar processes never receive the TERM signal that would allow graceful shutdown. This PR fixes that.

This PR was inspired by https://github.com/kafkaesque-io/pulsar-helm-chart/pull/31

### DIFF
--- a/charts/pulsar/templates/autorecovery-statefulset.yaml
+++ b/charts/pulsar/templates/autorecovery-statefulset.yaml
@@ -129,7 +129,7 @@ spec:
         - >
           bin/apply-config-from-env.py conf/bookkeeper.conf;
           {{- include "pulsar.autorecovery.zookeeper.tls.settings" . | nindent 10 }}
-          bin/bookkeeper autorecovery
+          exec bin/bookkeeper autorecovery
         ports:
         - name: http
           containerPort: {{ .Values.autorecovery.ports.http }}
--- a/charts/pulsar/templates/bookkeeper-statefulset.yaml
+++ b/charts/pulsar/templates/bookkeeper-statefulset.yaml
@@ -154,7 +154,7 @@ spec:
         - >
           bin/apply-config-from-env.py conf/bookkeeper.conf;
           {{- include "pulsar.bookkeeper.zookeeper.tls.settings" . | nindent 10 }}
-          bin/pulsar bookie;
+          exec bin/pulsar bookie;
         ports:
         - name: bookie
           containerPort: {{ .Values.bookkeeper.ports.bookie }}
--- a/charts/pulsar/templates/broker-statefulset.yaml
+++ b/charts/pulsar/templates/broker-statefulset.yaml
@@ -195,7 +195,7 @@ spec:
             bin/pulsar zookeeper-shell -server {{ template "pulsar.zookeeper.connect" . }} get {{ template "pulsar.broker.znode" . }};
           done;
           cat conf/pulsar_env.sh;
-          bin/pulsar broker;
+          exec bin/pulsar broker;
         ports:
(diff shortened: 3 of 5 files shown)


sijie:
@lhotari [bin/pulsar](https://github.com/apache/pulsar/blob/master/bin/pulsar#L314) script already uses `exec` to running the process. Are you sure that we need this change?

lhotari:
> @lhotari [bin/pulsar](https://github.com/apache/pulsar/blob/master/bin/pulsar#L314) script already uses `exec` to running the process. Are you sure that we need this change?

Yes, it's needed. `exec` is a shell built-in and it will only be able to replace it's current shell not any parent shells. Therefore the exec in `bin/pulsar` won't affect the top-level shell (`command: ["sh", "-c"]`).

You can quickly check the behavior without exec in docker:
```
docker run --name pulsar-standalone --rm -it --entrypoint '/bin/sh' apachepulsar/pulsar:2.6.1 -c 'bin/apply-config-from-env.py conf/standalone.conf && bin/pulsar standalone'
```

```
$ docker exec -it pulsar-standalone ps -ef
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  0 19:56 pts/0    00:00:00 /bin/sh -c bin/apply-config-from
root         7     1 99 19:56 pts/0    00:01:01 /usr/local/openjdk-8/bin/java -c
root       685     0 10 19:56 pts/1    00:00:00 ps -ef
```

As it can be seen, /bin/sh is PID 1 in the docker container when exec isn't used.


after adding `exec` this changes:
```
docker run --name pulsar-standalone --rm -it --entrypoint '/bin/sh' apachepulsar/pulsar:2.6.1 -c 'bin/apply-config-from-env.py conf/standalone.conf && exec bin/pulsar standalone'
```

```
$ docker exec -it pulsar-standalone ps -ef
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0 99 19:58 pts/0    00:00:14 /usr/local/openjdk-8/bin/java -c
root       253     0  0 19:58 pts/1    00:00:00 ps -ef
```

the `java` process is now PID 1 and it will receive OS signals as recommended in Docker and k8s documentation (links provided in the issue description).

sijie:
@lhotari Thank you for your detailed explanation! It makes sense to me.

PR #75, opened 2020-10-20
Local mode for kubernetes object generators

pinkavaj:
### Motivation

This allows operation in environemnts where direct installation of objects into
kubernetes cluster is not desired or possible. For example when using sealedsecrets
or SOPS, where the secrets are firs encrypted and then commited into repository
and deployed latter by some other deployment system.

### Modifications

 * allow select betwen instllation of generated kubernetes objects into cluster or write the into stadard output
 * write all user prompts into standar error output, so they does not mess with generated kubernetes objects

### Verifying this change

### DIFF
--- a/scripts/pulsar/generate_token.sh
+++ b/scripts/pulsar/generate_token.sh
@@ -32,6 +32,7 @@ Options:
        -k,--release                     the pulsar helm release name
        -r,--role                        the pulsar role
        -s,--symmetric                   use symmetric secret key for generating the token. If not provided, the private key of an asymmetric pair of keys is used.
+       -l,--local                       read and write output from local filesystem, do not install secret to kubernetes
 Usage:
     $0 --namespace pulsar --release pulsar-dev -c <pulsar-role>
 EOF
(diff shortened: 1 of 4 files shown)


sijie (approved):

scripts/pulsar/generate_token.sh
```diff
@@ -111,11 +127,11 @@ function pulsar::jwt::generate_asymmetric_token() {
     trap "test -f $privatekeytmpfile && rm $privatekeytmpfile" RETURN
     tokentmpfile=$(mktemp)
     trap "test -f $tokentmpfile && rm $tokentmpfile" RETURN
-    kubectl get -n ${namespace} secrets ${secret_name} -o jsonpath="{.data['PRIVATEKEY']}" | base64 --decode > ${privatekeytmpfile}
+    pulsar::jwt::get_secret SECRETKEY ${tmpfile}
```
jeantil, line 130: this should have been `${privatekeytmpfile}`, fixed in  #80 through 3185b14d5c97e24e9de02d0d07f7be4434d38f99

PR #87, opened 2020-11-24
Add basic PSP and RBAC for core components

miecio45:
Add PSP and add/modify RBAC. I'm open for all discussion.

### Motivation

On clusters which use PSP and restrictive default policy pulsar cannot be installed, because it uses root user and requires writable container root directory. Additionally default RBAC for broker are too permissive (usage of ClusterRoleBinding) in my opinion.

### Modifications

Add PSP and RBAC for bookkeeper and autorecovery to add
exception to allow startup even in secure environment
where containers cannot access RW on root by default.

Add option for limiting broker ClusterRoleBinding
to single namespace by replacing to RoleBinding

### Verifying this change

### DIFF
--- a/charts/pulsar/Chart.yaml
+++ b/charts/pulsar/Chart.yaml
@@ -21,7 +21,7 @@ apiVersion: v1
 appVersion: "2.7.0"
 description: Apache Pulsar Helm chart for Kubernetes
 name: pulsar
-version: 2.7.0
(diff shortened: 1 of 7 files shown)


miecio45:
I rebased to new master, CI is passing now. All comments are welcome :)

sijie (approved):

PR #96, opened 2021-01-21
Automatically pass bookieId to the bookie as metadata.name

eolivelli:
### Motivation

In Pulsar 2.7.x you can configured a bookieId that is the unique identifier of the Bookie and it does not depend on the network address of the bookie.
In Kubernetes we can set the bookieId as the name of the Pod, this way we are going to decouple the BookieId from the hostname and also from the main Bookie binary protocol endpoint (hostname:port)

see
http://bookkeeper.apache.org/bps/BP-41-bookieid/

### Modifications
Add new configuration parameter `bookkeeper.usePodNameAsBookieId`
in that case we pass the PULSAR_PREFIX_bookieId env variable, filled with metadata.name.

It is not possible to upgrade an existing Bookie to this new mode, in order to switch to BookieId manually you have to set  bookieId equals to the current bookieId you are seeing for the bookie, but this cannot be done within the Helm Chart.

### DIFF
--- a/charts/pulsar/templates/bookkeeper-statefulset.yaml
+++ b/charts/pulsar/templates/bookkeeper-statefulset.yaml
@@ -178,10 +178,12 @@ spec:
         - configMapRef:
             name: "{{ template "pulsar.fullname" . }}-{{ .Values.bookkeeper.component }}"
         env:
+        {{- if .Values.bookkeeper.usePodNameAsBookiedId }}
         - name: PULSAR_PREFIX_bookieId
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
+        {{- end}}
         volumeMounts:
         {{- if .Values.bookkeeper.volumes.useSingleCommonVolume }}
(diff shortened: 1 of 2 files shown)


sijie (changes requested):
Not sure we need bookieId change in Kubernetes because StatefulSet already provides a stable identifier. `bookieId` is useful when running directly on VM or AutoScalingGroup.

sijie:
Also it is unclear to me what is the impact and migration path here.

eolivelli:
Currently the bookie id generated by the bookie is computed from the main bookie protocol endpoint.
In the future we are going to add new endpoints and protocols (as in your SNI patch)
Having references to the current port is a burden we should take off.

In case of Kubernetes the pod name looks like a good and meaningful id.

Another option would be to let the bookie generate an UUID and store it locally on disk, as we are doing in HerdDB.

Regarding the upgrade story: this is the hard point and I still haven't found a good solution, this is why this PR is only a draft.

It would be good to have the ability to use the  BookieID  configured only for new bookies for instance. But I don't know if this is feasible with an helm chart.

lhotari:
Closing since there doesn't seem to be a need.

PR #107, opened 2021-03-03
makes cert-manager apiVersion configurable

### LINKED ISSUE #68: Cert-manager resources api-version is hardcoded
**Describe the bug**
Enabling `certs.internal_issuer.enabled=true` results in failed setup as CRD for cert-managers resources does not exist:
```
Error: unable to build kubernetes objects from release manifest: [unable to recognize "": no matches for kind "Certificate" in version "cert-manager.io/v1alpha2", unable to recognize "": no matches for kind "Issuer" in version "cert-manager.io/v1alpha2"]
```
cert-manager has released a stable version with updated `apiVersion:  cert-manager.io/v1`
fyi: there are many cert-manager api versions v1alpha2,v1alpha3,v1beta1

**To Reproduce**
helm upgrade --install pulsar -n pulsar apache/pulsar --values certs.internal_issuer.enabled=true

**Expected behavior**
Configurable external CRD api-versions in value.yaml

jeantil:
This commit let's users override the apiVersion referenced in this
chart so that the chart can be used with newer cert-manager releases.
(script/cert-manager/install-cert-manager.sh installs 0.13.0 when
current version is 1.2.0...)

Fixes #68

### Motivation

cert-manager apiVersion changed after cert-manager 1.0.0 was released, which prevents the chart from provisionning certificates with newer cert-manager installation because of an incompatible apiVersion.

I have a cluster with cert-manager >1.0.0 installed, making `apiVersion` overridable makes it easy for me to install pulsar on that cluster

### Modifications

I introduced the value `certs.internal_issuer.apiVersion`, which by default uses the apiVersion that was previously hardcoded (`cert-manager.io/v1alpha2`) 
I replaced all occurrences of that apiVersion by a reference to the value so that users can override it to `cert-manager.io/v1` if they have a newer version of cert-manager installed.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/tls-cert-internal-issuer.yaml
+++ b/charts/pulsar/templates/tls-cert-internal-issuer.yaml
@@ -19,7 +19,7 @@
 
 {{- if .Values.certs.internal_issuer.enabled }}
 {{- if eq .Values.certs.internal_issuer.type "selfsigning" }}
-apiVersion: cert-manager.io/v1alpha2
+apiVersion: "{{ .Values.certs.internal_issuer.apiVersion }}"
 kind: Issuer
 metadata:
   name: "{{ template "pulsar.fullname" . }}-{{ .Values.certs.internal_issuer.component }}"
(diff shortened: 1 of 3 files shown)


mbaechler:
Any chance we get that PR merged?

sijie (approved):

jeantil:
@sijie thank you !

PR #118, opened 2021-04-30
Consistently use the 'pulsar.matchLabels' template for matching

MMeent:
This also limits the scope of the PodMonitors to the resources of only this install, instead of all installs that share `component:` label values.

Fixes some issues with overly broad PodMonitor selectors, and put all selectors (except affinity selectors) in one line w.r.t. how they are generated.

### Motivation

This chart has several templates for generating standardized labels, which can be overridden by the user for using their own label scheme. Due to inconsistently applied labelSelector templates, this is only partially viable. By consistently applying the `pulsar.matchLabels` template, this makes the chart more extendable, and by extension, more useable. 

Additionally, PodMonitor resources specified an overly broad scope for monitoring components of the chart, where it would e.g. watch all `component: bookie` pods, even those not generated by this chart. This is also fixed with this pull request.

### Modifications

I've updated all labelSelector: -elements that I could find to use `pulsar.matchLabels` instead of hardcoded app + release labels.

Additionally, I've narrowed the scope of the various PodMonitor resources to only watch resources that should have been generated by this release, such that a PodMonitor for this chart does not also monitor another chart (as that would be extra load on the servers).

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/autorecovery-podmonitor.yaml
+++ b/charts/pulsar/templates/autorecovery-podmonitor.yaml
@@ -50,5 +50,6 @@ spec:
           targetLabel: kubernetes_pod_name
   selector:
     matchLabels:
+      {{- include "pulsar.matchLabels" . | nindent 6 }}
       component: {{ .Values.autorecovery.component }}
 {{- end }}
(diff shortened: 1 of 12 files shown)


lhotari (approved):
Good work @MMeent !

PR #127, opened 2021-06-10
Antiaffinity with zone topology for high availability 

sdwerwed:
### Motivation

There is no option in the current helm chart to make the pulsar high available. With this pull request, we can make some of the pulsar components highly available (multiAZ support). This is usually a requirement by many companies and I believe it will help many people that are facing this limitation. This version is backwards compatible, the default values are the same. It is tested and works as expected, pods can be scheduled in different availability zones if `<component>.affinity.anti_affinity_topology.zone.enabled: true`

### Modifications

I have added an extra option in the anti-affinity section for multiAZ support (disabled by default), this future is applied to bookkeeper, proxy, zookeeper, auto-recovery, and broker and is almost the same code for all of them. Made the weight from static value to variable (defaults are not changed), so users can adjust the weight in case they want to use both `host` and `zone` anti-affinity 

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/autorecovery-statefulset.yaml
+++ b/charts/pulsar/templates/autorecovery-statefulset.yaml
@@ -67,6 +67,7 @@ spec:
         podAntiAffinity:
           {{ if eq .Values.autorecovery.affinity.type "requiredDuringSchedulingIgnoredDuringExecution"}}
           {{ .Values.autorecovery.affinity.type }}:
+          {{- if .Values.autorecovery.affinity.anti_affinity_topology.host.enabled}}
           - labelSelector:
               matchExpressions:
               - key: "app"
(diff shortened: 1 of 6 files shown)


sdwerwed:
As I am a first-time contributor, can a maintainer run the pipelines, please?

sdwerwed:
Hello @wolfstudy, @sijie, can you have a look at this PR, let me know if any other action is required from my side.

lhotari (approved):
Good work @sdwerwed . Thank you for the contribution.

sijie:
@sdwerwed: I think the zone affinity would work for stateless components. I don't think it will work well for stateful components. Because the location of bookie and zookeeper pods will be bound to the location of EBS volumes. Also, even you are able to place bookies across different zones. But you don't configure the rack information, so bookkeeper can potentially place two same copies in the pods running in the same zone.

Hence I believe the solution is incomplete. If we merge this, it will cause confusion to people and let them believe it is a "complete" solution.

sijie (changes requested):
would love to see the comment is addressed first.

michaeljmarshall:
> @sdwerwed: I think the zone affinity would work for stateless components. I don't think it will work well for stateful components. Because the location of bookie and zookeeper pods will be bound to the location of EBS volumes.

To add some AWS context, EBS volumes are restricted to a single availability zone within a region. Once a bookie gets an EBS volume, it cannot be migrated to another AZ. I haven't looked closely at this PR, but we'll need to make sure that it works for upgrading existing clusters that do not comply with the anti-affinity policy.

> Also, even you are able to place bookies across different zones. But you don't configure the rack information, so bookkeeper can potentially place two same copies in the pods running in the same zone. 

I agree that this could be misleading. We could mitigate the shortcoming with documentation, though. Also, this feature can work without configuring rack awareness. For example, if your ack quorum is greater than the number of bookies in a single AZ, you'll know that your writes are spread across multiple AZs and are therefore HA. This isn't as good as configuring rack awareness, but it is probably better than random bookie scheduling.

sdwerwed:
@sijie 

> I think the zone affinity would work for stateless components. I don't think it will work well for stateful components. Because the location of bookie and zookeeper pods will be bound to the location of EBS volumes. 

This is a known limitation of AWS and probably Azure too, however, the default value is not multi az, if an enginner would like to make it multi az can do it manually with additional steps, not sure if we can handle that in the helm chart anyway, we could document that or if you have another idea how we could make it fully functional with helm let me know.

> But you don't configure the rack information, so bookkeeper can potentially place two same copies in the pods running in the same zone.

I think I miss some context here as I did not get what is the issue, what do you mean to configure rack information? Could you give me some sources to read about it so I can suggest a solution?

frankjkelly:
Any update on this? Would be great to help make Pulsar HA

sdwerwed:
Closing this as got very old and need maintenance

PR #130, opened 2021-06-15
Bump pulsar 2.8.0

codelipenghui:


zymap (approved):
LGTM

Anonymitaet (approved):

jiazhai (approved):

Jennifer88huang (approved):

jiazhai (approved):

eolivelli (changes requested):
I am not sure we can upgrade the helm chart given the amount of problems that we are seeing.
We can work hard to help BK community to release BK 4.14.2 as soon as possible for instance.

There is no hurry to release an helm chart.
If people want to upgrade they can use this patch as sample but we should not provide a chart with tricks like using an older version of Pulsar for some components.

It will create only confusion for the users 

@merlimat @sijie @rdhabalia what so you think?

mkoertgen:
> BK 4.14.2

Not 100% sure but BK 4.14.3 was released 3w ago: https://github.com/apache/bookkeeper/releases/tag/release-4.14.3

We are looking forward to use the function-package-service introduced in [2.8.0](https://pulsar.apache.org/release-notes/#2.8.0) and would love to upgrade our pulsar clusters to 2.8+.

Carmezim:
Hey guys, like @mkoertgen pointed out BK `v4.12.2` has been added to pulsar `v2.8.0` release, is there any other blockers to merge that PR and upgrade the chart?

cc @eolivelli

lhotari:
superseded by #190

charts/pulsar/values.yaml
```diff
@@ -278,7 +278,7 @@ zookeeper:
   replicaCount: 3
   updateStrategy:
     type: RollingUpdate
-  podManagementPolicy: OrderedReady
+  podManagementPolicy: Parallel
```
lhotari, line 281: @codelipenghui Do we need this change for #190 ? Can you remember the reason for this change? /cc @michaeljmarshall @315157973

michaeljmarshall: I think `Parallel` is the right choice for us. It'll lead to faster cluster initialization. https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#parallel-pod-management

PR #138, opened 2021-07-15
automate initialize

### LINKED ISSUE #114: Use hooks instead of initialize=true
Why don't use pre-install or post-install hooks?
Any reason?

https://helm.sh/docs/topics/charts_hooks/
115307963-4d467300-a16a-11eb-83c3-8138ad2aaac2
115307997-5b948f00-a16a-11eb-8e33-e29da243fd17

https://github.com/apache/pulsar-helm-chart/pull/53

valeriano-manassero:
Fixes #114

### Motivation

Having `initialize` value that must be manually set is prone to human error, there is a risk of running init again forcing the need of a rollback.

### Modifications

Using Helm capability to understand if it's an install or upgrade we can avoid to launch init when not needed.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/bookkeeper-cluster-initialize.yaml
+++ b/charts/pulsar/templates/bookkeeper-cluster-initialize.yaml
@@ -16,7 +16,7 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-{{- if .Release.IsInstall }}
+{{- if or .Release.IsInstall .Values.initialize }}
 {{- if .Values.components.bookkeeper }}
 apiVersion: batch/v1
 kind: Job
--- a/charts/pulsar/templates/bookkeeper-cluster-initialize.yaml
+++ b/charts/pulsar/templates/bookkeeper-cluster-initialize.yaml
@@ -16,7 +16,7 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-{{- if .Values.initialize }}
+{{- if .Release.IsInstall }}
 {{- if .Values.components.bookkeeper }}
 apiVersion: batch/v1
 kind: Job
--- a/charts/pulsar/templates/pulsar-cluster-initialize.yaml
+++ b/charts/pulsar/templates/pulsar-cluster-initialize.yaml
@@ -17,7 +17,7 @@
 # under the License.
 #
 
-{{- if .Release.IsInstall }}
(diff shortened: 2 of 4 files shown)


sijie (approved):

valeriano-manassero:
This is going to be merged?

charts/pulsar/templates/bookkeeper-cluster-initialize.yaml
```diff
@@ -16,7 +16,7 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-{{- if .Values.initialize }}
+{{- if .Release.IsInstall }}
```
lhotari, line 19: I'd suggest keeping backwards compatibility. 
```suggestion
{{- if or .Release.IsInstall .Values.initialize }}
```
The benefit of this is that the metadata initialization job can be created on demand when needed. I have been using this in test environments. It's possible to clear the persistent state of the cluster by scaling stateful sets to 0, deleting PVCs and doing a deployment with `--set initialize=true`.

charts/pulsar/templates/pulsar-cluster-initialize.yaml
```diff
@@ -17,7 +17,7 @@
 # under the License.
 #
 
-{{- if .Values.initialize }}
+{{- if .Release.IsInstall }}
```
lhotari, line 20:
```suggestion
{{- if or .Release.IsInstall .Values.initialize }}
```

charts/pulsar/values.yaml
```diff
@@ -34,9 +34,6 @@ clusterDomain: cluster.local
 ### Global Settings
 ###
 
-## Set to true on install
```
lhotari, line 37: please revert this change to keep backwards compatibility.

lhotari (changes requested):

lhotari (approved):
LGTM

lhotari:
@valeriano-manassero Thank you for the contribution. Would it be also possible to update README.md in this PR and remove the instructions of using `--set initialize=true` when doing the installation?

valeriano-manassero:
> @valeriano-manassero Thank you for the contribution. Would it be also possible to update README.md in this PR and remove the instructions of using `--set initialize=true` when doing the installation?

done

lhotari (approved):

PR #140, opened 2021-07-29
Add Support for imagePullSecrets

### LINKED ISSUE #125: Append Configuration of ImagePullSecret
Currently it is difficult to pull dockerhub images if not logged in.

In chart's "values.yaml", append filed to set imagepullsecret.

toneill818:
Fixes #125

### Motivation

The default images in the values.yaml are in docker hub. This PR allows us to provide image pull secrets for the containers which will allow us to get around Docker Hub's rate limiting if the nodes are not logged into Docker Hub.

### Modifications

Added a new template to generate `imagePullSecrets`, and included them in the deployments and statefulsets. This will only add them if they are specified under `images.imagePullSecrets`

### Verifying this change

- [] Make sure that the change passes the CI checks.


charts/pulsar/templates/_helpers.tpl
```diff
@@ -84,3 +84,15 @@ Create the match labels.
 app: {{ template "pulsar.name" . }}
 release: {{ .Release.Name }}
 {{- end }}
+
+{{/*
+Create ImagePullSecrets
+*/}}
+{{- define "imagePullSecrets" -}}
```
michaeljmarshall, line 91: This template name is global. Even though this template is generic, I think we should namespace it with the helm chart's name. It could instead be `pulsar.imagePullSecrets`.

Here is a reference to the helm [documentation](https://helm.sh/docs/chart_template_guide/named_templates/#declaring-and-using-templates-with-define-and-template) for this convention:

> As mentioned above, template names are global. As a result of this, if two templates are declared with the same name the last occurrence will be the one that is used. Since templates in subcharts are compiled together with top-level templates, it is best to name your templates with chart specific names. A popular naming convention is to prefix each defined template with the name of the chart: {{ define "mychart.labels" }}.

michaeljmarshall (changes requested):
@toneill818 - left one comment on the helm convention for naming templates. Otherwise, this looks good to me.

toneill818:
@michaeljmarshall Are there any other changes I should make?

michaeljmarshall:
@toneill818 - did you test this PR with a value in the `.Values.images.imagePullSecrets`? The current tests wouldn't have covered that case, and your initial comment only mentions relying on the CI checks.

toneill818:
@michaeljmarshall Yes, I am currently using my branch with imagePullSecrets populated to work with Docker Hub's rate limits.

michaeljmarshall (approved):
Thanks for confirming. LGTM.

michaeljmarshall:
@sijie - PTAL

sijie (approved):

PR #150, opened 2021-08-17
added additionalCommand parameter

### LINKED ISSUE #148: Add the ability to specify a custom command to the start up process of statefulset pods
We would like to be able to run the `update-ca-certificates` command on start-up of pods

Pods should be able to have a custom command added to the `command:` line

We've tried baking our own images to work around this limitation but it would be better if the support was in the helm chart directly

acjohnson:
Fixes #148 

### Motivation
Added a `additionalCommand` parameter for each statefulset so that the chart user can optionally specific a custom command such as `update-ca-certificates` to run prior to starting the JVM and other container processes. This is required to avoid custom image builds simply for shipping an internal CA certificate file.

### Modifications
* Added `additionalCommand` parameter to values.yaml for each statefulset
* Added conditionals to each statefulset allowing the chart user to specify a command to be ran at the beginning of the command line(s)

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/bookkeeper-statefulset.yaml
+++ b/charts/pulsar/templates/bookkeeper-statefulset.yaml
@@ -165,6 +165,9 @@ spec:
         command: ["sh", "-c"]
         args:
         - >
+        {{- if .Values.bookkeeper.additionalCommand }}
+          {{ .Values.bookkeeper.additionalCommand }}
+        {{- end }}
           bin/apply-config-from-env.py conf/bookkeeper.conf;
           {{- include "pulsar.bookkeeper.zookeeper.tls.settings" . | nindent 10 }}
           OPTS="${OPTS} -Dlog4j2.formatMsgNoLookups=true" exec bin/pulsar bookie;
--- a/charts/pulsar/templates/broker-statefulset.yaml
+++ b/charts/pulsar/templates/broker-statefulset.yaml
@@ -198,6 +198,9 @@ spec:
         command: ["sh", "-c"]
         args:
         - >
+        {{- if .Values.broker.additionalCommand }}
+          {{ .Values.broker.additionalCommand }}
+        {{- end }}
           bin/apply-config-from-env.py conf/broker.conf;
           bin/gen-yml-from-env.py conf/functions_worker.yml;
           echo "OK" > status;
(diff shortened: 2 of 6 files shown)


charts/pulsar/templates/bookkeeper-statefulset.yaml
```diff
@@ -162,6 +162,9 @@ spec:
         command: ["sh", "-c"]
         args:
         - >
+        {{- if .Values.bookkeeper.additionalCommand }}
+          {{ .Values.bookkeeper.additionalCommand }}
+        {{- end }}
```
michaeljmarshall, line 167: Did you verify that this renders properly using `helm template`? I copied this into my repo and seem to be getting an error.

michaeljmarshall (commented):
Thank you for your contribution, @acjohnson. I am not opposed to adding a value like this, but I should point out that we are going to modify the pulsar docker image so that it runs with a non-root user by default. This work hasn't completed yet. When it does, a command like `update-ca-certificates` will fail because it requires elevated permission.

charts/pulsar/templates/bookkeeper-statefulset.yaml
acjohnson: Yes it renders properly for me @michaeljmarshall. I'm running Helm v3.6.3

```shell
# helm version --short
v3.6.3+gd506314

# helm template pulsar/ | tail
# software distributed under the License is distributed on an
# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
# KIND, either express or implied.  See the License for the
# under the License.
#

# deploy zookeeper only when `components.zookeeper` is true

# define the storage class for data directory

# echo $?
0
```

acjohnson:
> Thank you for your contribution, @acjohnson. I am not opposed to adding a value like this, but I should point out that we are going to modify the pulsar docker image so that it runs with a non-root user by default. This work hasn't completed yet. When it does, a command like `update-ca-certificates` will fail because it requires elevated permission.

Hmm... yeah that's too bad it won't work after that non-root change...

lhotari (approved):
LGTM

michaeljmarshall (approved):
LGTM.

michaeljmarshall:
This PR adds value, even though the original goal of this feature might break when we move to a non-root docker image, so I am going to merge it now.

PR #152, opened 2021-08-24
Define configuration store settings on proxy configmap

josephglanville:
### Motivation

This is required to active websockets and while it can be provided via
proxy.configData this is more ergonomic for users of the chart.

### Modifications

Added zookeeperServers/configurationStoreServers to proxy config map

### Verifying this change


josephglanville:
@codelipenghui was this closed accidentally?

frankjkelly:
@josephglanville that was my screw-up - apologies. @codelipenghui can you help? Thanks!

PR #155, opened 2021-09-01
Add grafana data persistence support.

### LINKED ISSUE #154: Grafana data lost once pod restart.
Now, pulsar-helm-chart use empty dir instead of pv. So once grafana pod restarts, the grafana data lost.

Add grafana data persistence support.

dragonls:
Fixes #154 

### Motivation

Add pulsar pvc to support grafana data persistence.

### Modifications
- Add `grafana-pvc.yaml`.
- Add grafana data mount logic in `grafana-deployment.yaml`
- Add `grafana.volumns` option in `values.yaml` to control the grafana data volumes.

### Verifying this change


charts/pulsar/templates/grafana-pvc.yaml
```diff
@@ -0,0 +1,40 @@
...
+  namespace: {{ template "pulsar.namespace" . }}
+spec:
+  resources:
+    requests:
+      storage: {{ .Values.grafana.volumes.data.size }}
+  accessModes: [ "ReadWriteOnce" ]
+{{- if .Values.grafana.volumes.data.storageClassName }}
+  storageClassName: "{{ .Values.grafana.volumes.data.storageClassName }}"
+{{- else if and (not (and .Values.volumes.local_storage .Values.grafana.volumes.data.local_storage)) .Values.grafana.volumes.data.storageClass }}
+  storageClassName: "{{ template "pulsar.fullname" . }}-{{ .Values.grafana.component }}-{{ .Values.grafana.volumes.data.name }}"
```
MarvinCai, line 35: other components create their own storageclass so we can use this storageclass, seem that's not true with grafana here, then we should either remove this or add creation for storageclass.

dragonls: Add creation for grafana storageclass.

charts/pulsar/templates/grafana-deployment.yaml
```diff
@@ -88,4 +91,20 @@ spec:
               name: "{{ template "pulsar.fullname" . }}-{{ .Values.grafana.component }}-secret"
               key: GRAFANA_ADMIN_PASSWORD
         {{- include "pulsar.imagePullSecrets" . | nindent 6}}
+      volumes:
+        {{- if not (and (and .Values.persistence .Values.volumes.persistence) .Values.grafana.volumes.persistence) }}
```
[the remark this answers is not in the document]
dragonls: Simplified.

dragonls:
Sorry for the late modification. @MarvinCai PTAL.

charts/pulsar/values.yaml
```diff
@@ -988,6 +988,24 @@ grafana:
     requests:
       memory: 250Mi
       cpu: 0.1
+  volumes:
+    # use a persistent volume or emptyDir
+    persistence: true
+    data:
+      name: data
+      size: 10Gi
```
lhotari, line 996: 10Gi seems a lot for grafana persistence. I'd assume that 100Mi would be sufficient.

charts/pulsar/values.yaml
```diff
@@ -988,6 +988,24 @@ grafana:
     requests:
       memory: 250Mi
       cpu: 0.1
+  volumes:
+    # use a persistent volume or emptyDir
+    persistence: true
```
michaeljmarshall, line 993: What is the reasoning for making this the default behavior? Alternatively, we could mount dashboards as config maps and avoid the extra storage costs associated with a PVC.

dragonls: Just thinking that user might be confused if the grafana pod restarts and data lost, once user modified the grafana data such as the dashboards.

charts/pulsar/values.yaml
dragonls: Updated.

charts/pulsar/values.yaml
michaeljmarshall: I agree there is a chance for confusion. However, the official Grafana helm chart disables persistence by default, so I think we should too.

https://github.com/grafana/helm-charts/blob/2a422f6c43c3db33d31fc12d23957d5b6ff5e8c6/charts/grafana/values.yaml#L275-L277

dragonls: I agree, change default persistence value to false.

PR #159, opened 2021-09-03
Fix #152 Helm chart does not support Istio port naming

frankjkelly:
Fixes #152 

### Motivation

Change port naming to abide by Istio protocol rules

### Modifications

Changed port names
- pulsar -> tcp-pulsar
- pulsarssl --> tls-pulsarssl etc

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/bookkeeper-service.yaml
+++ b/charts/pulsar/templates/bookkeeper-service.yaml
@@ -33,7 +33,7 @@ metadata:
 spec:
   ports:
   - name: bookie
-    port: {{ .Values.bookkeeper.ports.bookie }}
+    port: {{ .Values.bookkeeper.ports.tcp-bookie }}
   - name: http
     port: {{ .Values.bookkeeper.ports.http }}
   clusterIP: None
--- a/charts/pulsar/Chart.yaml
+++ b/charts/pulsar/Chart.yaml
@@ -21,7 +21,7 @@ apiVersion: v1
 appVersion: "2.7.2"
 description: Apache Pulsar Helm chart for Kubernetes
 name: pulsar
-version: 2.7.2
+version: 2.7.2.1
 home: https://pulsar.apache.org
 sources:
 - https://github.com/apache/pulsar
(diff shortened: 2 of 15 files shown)


charts/pulsar/templates/bookkeeper-service.yaml
```diff
@@ -33,7 +33,7 @@ metadata:
 spec:
   ports:
   - name: bookie
-    port: {{ .Values.bookkeeper.ports.bookie }}
```
lhotari, line 36: It seems that the keys used in the values should be kept unchanged since it would break backwards compatibility and that's probably also the source of linting failures?

frankjkelly: @lhotari thanks for the feedback. @addisonj or @sijie Lari raises a great point - is there another way around this? 
Istio also supports setting the `appProtocol` but requires Kubernetes 1.18+
https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection

sijie: I don't think it breaks any compatibility. Because it can easily upgrade and roll back.

An alternative approach is to add a "portPrefix". By default it is `""`. But users can override it with "tcp-”.

lhotari: It seems that my comment about the keys and backwards compatibility needs clarification. For changing the port names, it shouldn't be necessary to change the content of the values yaml format at all. That's what my original comment was about.

lhotari: I like the portPrefix solution proposed by @sijie. (perhaps it could be called `portNamePrefix` for better clarity)

lhotari: well it seems that it's not "tcp-" for all ports, so perhaps it needs revisiting. in some cases, the prefix is "tls-"

frankjkelly: Thanks @sijie @lhotari - great idea of some prefixes that by default are blank for backwards compatibility- I'll try to PR that up today

frankjkelly:
FYI closing this in favor of another PR  #162

PR #165, opened 2021-10-09
useV2WireProtocol for bookkeeper autorecovery

hangc0276:
### Motivation
Currently we only set `bookkeeperUseV2WireProtocol=true` on Pulsar brokers to enforce brokers to use v2 protocol to avoid GC issues introduced by protobuf. 

However, we don’t set this setting on autorecovery, and the default value is `false`, Hence AutoRecovery will use v3 protocol.

We'd better set `useV2WireProtocol=true` in autorecovery in the helm charts

### Modifications

Set `useV2WireProtocol=true` in autorecovery.

### Verifying this change

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -597,6 +597,7 @@ autorecovery:
   configData:
     BOOKIE_MEM: >
       -Xms64m -Xmx64m
+    PULSAR_PREFIX_useV2WireProtocol: "true"
 
 ## Pulsar Zookeeper metadata. The metadata will be deployed as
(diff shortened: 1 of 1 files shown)


frankjkelly:
Just for my information can someone explain why Autorecovery should use the V2 protocol but the Broker V3?  Thanks in advance!

hangc0276:
> Just for my information can someone explain why Autorecovery should use the V2 protocol but the Broker V3? Thanks in advance!

@frankjkelly I broker side, it has been set the bookie client use V2 protocol by default.

lhotari (approved):

PR #169, opened 2021-11-04
When K8S does not support PVC, mount the storage to local disk.

autumnqfeng:
Fixes #<xyz>

### Motivation

Installing ZooKeeper and BookKeeper through Helm Charts fails when K8S does not support PVC, so I think Helm Charts should support mounting to local disks.

### Modifications

* `values.yaml` added the `hostPath` configuration for mounting to a local disk.
* Mount the `/zookeeper/data` directory to the local disk in  `zookeeper-statefulset.yaml` .
* In `bookkeeper-statefulset.yaml`, mount the `/bookie/journal` and `/bookie/ledgers` to the local disk. For example, mount journal to the local SSD and ledgers to HDD.

### Verifying this change

- [ ]  Make sure that the change passes the CI checks.


lhotari:
Closing this since it doesn't seem to be common anymore that PVC isn't supported. If there's more interest for adding this support, we can reconsider later.

PR #176, opened 2021-11-22
Fix cluster initialize blocked when fail: #175

### LINKED ISSUE #175: Create may block if the initialize cluster job fails
**Describe the bug**
In the pulsar-init Job, the init script is:

``` shell
bin/pulsar initialize-cluster-metadata \
                --cluster pulsar \
                --zookeeper pulsar-zookeeper:2181 \
                --configuration-store pulsar-zookeeper:2181 \
                --web-service-url http://pulsar-broker.test.svc.cluster.local:8080/ \
                --web-service-url-tls https://pulsar-broker.test.svc.cluster.local:8443/ \
                --broker-service-url pulsar://pulsar-broker.test.svc.cluster.local:6650/ \
                --broker-service-url-tls pulsar+ssl://pulsar-broker.test.svc.cluster.local:6651/ || true;
```

The `|| true` at the end makes the Job not fail even when error occurs in initializing metadata.

Usually it's ok, for there's a init container checking zookeeper status, but if we have a bad day when zookeeper goes down just after init container succeeded, and before main container start, the cluster metadata will never be initialized. Which will result in pulsar components blocked forever. like in issue: #124 

We should remove the tail || true, let the job restart if failed.

**To Reproduce**
`TBD`

**Expected behavior**
init Job will keep retrying when initializing metadata failed, until success.

**Screenshots**
`TBD`

 - OS: KinD on OS.X

**Additional context**
ref: milvus-io/milvus-operator#40

haorenfsa:
### Motivation

Fixes #175 

### Modifications

Cluster init job retries every 3 sec before success

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/pulsar-cluster-initialize.yaml
+++ b/charts/pulsar/templates/pulsar-cluster-initialize.yaml
@@ -100,11 +100,11 @@ spec:
               --web-service-url http://{{ template "pulsar.fullname" . }}-{{ .Values.broker.component }}.{{ template "pulsar.namespace" . }}.svc.{{ .Values.clusterDomain }}:{{ .Values.broker.ports.http }}/ \
               --web-service-url-tls https://{{ template "pulsar.fullname" . }}-{{ .Values.broker.component }}.{{ template "pulsar.namespace" . }}.svc.{{ .Values.clusterDomain }}:{{ .Values.broker.ports.https }}/ \
               --broker-service-url pulsar://{{ template "pulsar.fullname" . }}-{{ .Values.broker.component }}.{{ template "pulsar.namespace" . }}.svc.{{ .Values.clusterDomain }}:{{ .Values.broker.ports.pulsar }}/ \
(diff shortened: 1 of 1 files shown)


MarvinCai (approved):
lgtm

haorenfsa:
@eolivelli  have a look?

lhotari (changes requested):
Instead of adding the loop to retry forever, I'd suggest to remove the `|| true` and change `restartPolicy: Never` to `restartPolicy: OnFailure`.

michaeljmarshall (dismissed):
I agree with @lhotari. Let's rely on Kubernetes to restart the pod. That'll also give us an increasing back off for free.

haorenfsa:
updated

lhotari (approved):
LGTM

frankjkelly:
@lhotari  without a change in Chart version will a new release be created? CC: @sijie

lhotari:
> @lhotari without a change in Chart version will a new release be created? CC: @sijie

correct. Here's the logs: https://github.com/apache/pulsar-helm-chart/runs/4715446366?check_suite_focus=true .
I think it's useful to have multiple PRs batched into a release.

frankjkelly:
@lhotari Sounds good, thanks for the clarification.

michaeljmarshall:
>  I think it's useful to have multiple PRs batched into a release.

+1

PR #179, opened 2021-11-25
Make k8s probe timeoutSeconds configurable and fix k8s 1.20+ compatibility

lhotari:
### Motivation

- address compatibility with Kubernetes 1.20+. This impacts "bin/pulsar-zookeeper-ruok.sh" exec probe used in ZK.
  "Before Kubernetes 1.20, the field timeoutSeconds was not respected for exec probes: probes continued running indefinitely, even past their configured deadline, until a result was returned."
   https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes
- make k8s probe timeoutSeconds configurable

### Modifications

- Add timeoutSeconds to startup, readiness and liveness probes
- Set the default to 5 seconds


eolivelli (approved):
+1

PR #183, opened 2021-12-02
Update ingress api version, extension/v1beta1 will not be supported i…

### LINKED ISSUE #157: Ingresses still use the extensions/v1beta1 versioned resource
**Describe the bug**
All the ingress templates still use the deprecated (since Kubernetes 1.19) extensions/v1beta1 Ingress instead of networking.k8s.io/v1 Ingress. In addition, support for this type will be dropped in the upcoming major Kubernetes version (1.22)

**To Reproduce**
1. Take the latest helm chart
2. Apply the helm chart with some ingresses turned on
3. Notice the warnings that extensions/v1beta1 Ingress is deprecated and support will be dropped in 1.22

**Expected behavior**
The chart should use the new networking.k8s.io/v1 Ingress.

wangshu3000:
…n new k8s version, this change keep backward compatibility for lower kubernetes version 

Fixes #157

### Motivation

The ingress api extensions/v1beta1 version will not be supported in k8s version 1.22+, need to update to new networking.k8s.io/v1.
This PR also keep support for old k8s version, to provide enough compatibility.

### Modifications

Updated the api version for below 4 components:
1. dashboard-ingress
2. grafana-ingress
3. proxy-ingress
4. pulsar-manager-ingress

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/dashboard-ingress.yaml
+++ b/charts/pulsar/templates/dashboard-ingress.yaml
@@ -19,7 +19,11 @@
 
 {{- if .Values.extra.dashboard }}
 {{- if .Values.dashboard.ingress.enabled }}
+{{- if semverCompare "<1.19-0" .Capabilities.KubeVersion.GitVersion }}
 apiVersion: extensions/v1beta1
+{{- else }}
+apiVersion: networking.k8s.io/v1
+{{- end }}
 kind: Ingress
 metadata:
   labels:
@@ -45,8 +49,17 @@ spec:
       http:
         paths:
           - path: {{ .Values.dashboard.ingress.path }}
+            {{- if semverCompare "<1.19-0" .Capabilities.KubeVersion.GitVersion }}
             backend:
               serviceName: "{{ template "pulsar.fullname" . }}-{{ .Values.dashboard.component }}"
               servicePort: {{ .Values.dashboard.ingress.port }}
+            {{- else }}
+            pathType: ImplementationSpecific
+            backend:
+              service:
+                name: "{{ template "pulsar.fullname" . }}-{{ .Values.dashboard.component }}"
+                port:
+                  number: {{ .Values.dashboard.ingress.port }}
+            {{- end }}
 {{- end }}
 {{- end }}
--- a/charts/pulsar/templates/dashboard-ingress.yaml
+++ b/charts/pulsar/templates/dashboard-ingress.yaml
@@ -19,7 +19,7 @@
 
 {{- if .Values.extra.dashboard }}
 {{- if .Values.dashboard.ingress.enabled }}
-{{- if semverCompare "<1.19-0" .Capabilities.KubeVersion.GitVersion }}
+{{- if semverCompare "<1.19-0" .Capabilities.KubeVersion.Version }}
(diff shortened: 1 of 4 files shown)


wangshu3000:
Could anybody in the team please review this PR? Thanks.

lhotari (approved):
Good work!

charts/pulsar/templates/dashboard-ingress.yaml
```diff
@@ -19,7 +19,11 @@
 
 {{- if .Values.extra.dashboard }}
 {{- if .Values.dashboard.ingress.enabled }}
+{{- if semverCompare "<1.19-0" .Capabilities.KubeVersion.GitVersion }}
```
michaeljmarshall, line 22: @wangshu3000 - do you have a reference that documents `GitVersion`? I don't see that in the helm documentation (https://helm.sh/docs/chart_template_guide/builtin_objects/).

I think we should prefer the logic described in this stack overflow answer: https://stackoverflow.com/a/66191437. Essentially, we can inspect the `APIVersion` to see if the api has `networking.k8s.io/v1` and branch accordingly.

michaeljmarshall (changes requested):
Great work @wangshu3000! I think it might make sense to modify the conditionals in the PR. Let me know what you think. Otherwise, this looks good to me.

michaeljmarshall:
@lhotari - do you know why the CI checks wouldn't have run for this PR? I haven't seen that happen before.

charts/pulsar/templates/dashboard-ingress.yaml
wangshu3000: @michaeljmarshall  Thanks for the feedback. 
Looks like they changed the doc. I saw this from some examples of helm repo. Looks like it's been deprecated. The new value should be Capabilities.KubeVersion.Version

There is one problem with .Capabilities.APIVersions.Has
When someone use helm template command. It'll lock the kubernetes version to latest version in helm. So if they use new helm version 3.7+, there would always has the new api version even already specify the old kubenetes version in parameter. So the generated template yaml will have the new api version, when applying those yaml files, they will be failed in old kubernetes cluster. 

I think maybe we should change to KubeVersion.Version as the KubeVersion.GitVersion has been deprecated.

Please let me know if this make sense. 

Thanks.

wangshu3000: https://github.com/argoproj/argo-cd/issues/3594
This is the issue for your reference.
There are some discussion around this.

michaeljmarshall: @wangshu3000 that definitely makes sense. I didn't know about the feature's shortcoming, thanks for sharing. Otherwise, do you need to update it to `KubeVersion.GitVersion` to `KubeVersion.Version`? We also need to figure out why CI hasn't run for this PR.

wangshu3000: @michaeljmarshall  ok. let me do it.

michaeljmarshall (approved):
LGTM. @wangshu3000 - looks like the tests ran against this latest PR, so that's good too.

PR #186, opened 2021-12-10
[Security] Workaround for CVE-2021-44228 Log4J RCE when Log4J >= 2.10.0

lhotari:
### Motivation

CVE-2021-44228 , a severe RCE for Log4J.

The workaround is to set `-Dlog4j2.formatMsgNoLookups=true` system property. This workaround applies for Log4J >= 2.10.0 . This covers all vulnerable Pulsar versions since Log4J2 version was 2.10.0 when the dependency was introduced to Pulsar in https://github.com/apache/pulsar/pull/680 .

### Modifications

Add `OPTS="${OPTS} -Dlog4j2.formatMsgNoLookups=true"` prefix to calls to `exec bin/pulsar` and `exec bin/bookkeeper` . This results in `-Dlog4j2.formatMsgNoLookups=true` system property getting set.

### DIFF
--- a/charts/pulsar/templates/autorecovery-statefulset.yaml
+++ b/charts/pulsar/templates/autorecovery-statefulset.yaml
@@ -139,7 +139,7 @@ spec:
         - >
           bin/apply-config-from-env.py conf/bookkeeper.conf;
           {{- include "pulsar.autorecovery.zookeeper.tls.settings" . | nindent 10 }}
-          exec bin/bookkeeper autorecovery
+          OPTS="${OPTS} -Dlog4j2.formatMsgNoLookups=true" exec bin/bookkeeper autorecovery
         ports:
         - name: http
           containerPort: {{ .Values.autorecovery.ports.http }}
--- a/charts/pulsar/Chart.yaml
+++ b/charts/pulsar/Chart.yaml
@@ -21,7 +21,7 @@ apiVersion: v1
 appVersion: "2.7.2"
 description: Apache Pulsar Helm chart for Kubernetes
 name: pulsar
-version: 2.7.5
+version: 2.7.6
 home: https://pulsar.apache.org
 sources:
 - https://github.com/apache/pulsar
--- a/charts/pulsar/templates/bookkeeper-statefulset.yaml
+++ b/charts/pulsar/templates/bookkeeper-statefulset.yaml
@@ -167,7 +167,7 @@ spec:
         - >
           bin/apply-config-from-env.py conf/bookkeeper.conf;
           {{- include "pulsar.bookkeeper.zookeeper.tls.settings" . | nindent 10 }}
-          exec bin/pulsar bookie;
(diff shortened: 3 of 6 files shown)


frankjkelly:
I could be wrong but I think you will need to bump the Chart version https://github.com/apache/pulsar-helm-chart/blob/master/charts/pulsar/Chart.yaml#L24 ?

lhotari:
> I could be wrong but I think you will need to bump the Chart version https://github.com/apache/pulsar-helm-chart/blob/master/charts/pulsar/Chart.yaml#L24 ? Otherwise a new helm chart release won't be created.

Thanks @frankjkelly . I bumped the version.

lhotari:
Notice! This workaround doesn't apply for Pulsar Functions that use process runtime or k8s runtime. Pulsar Proxy, Broker, Bookie and Zookeeper components are covered, but Pulsar Functions run in a separate process won't pick up the flag.

eolivelli (approved):
I am +1 with this change.
IIUC there is no way to apply the system property in another way

thanks for doing this.
can you please start a discussion to cut a new release for the helm chart ?

lhotari:
I'm looking into ways how to also cover Pulsar Functions since the current changes in the PR don't cover that. Since it's hard to set the system property for a Pulsar Function Java instance, there's a possibility to patch the java-instance.jar file and add a file called adding a file `log4j2.component.properties` with a single line `log4j2.formatMsgNoLookups=true` to turn off the vulnerable feature.

lhotari:
It seems that Pulsar Functions cannot be covered with the Helm Chart changes alone when using Process Runtime or K8S Runtime. For Pulsar Functions, it's easier to patch a docker image and upgrade the Log4J version. Another option is to modify the /pulsar/instances/java-instance.jar and add a file `log4j2.component.properties` with `log4j2.formatMsgNoLookups=true`.

lhotari:
Here's an additional workaround for patching existing Docker images with an additional overlay which upgrades Log4J to 2.15.0: https://github.com/lhotari/pulsar-docker-images-patch-CVE-2021-44228 . That could be used to cover Pulsar Functions.

charts/pulsar/templates/autorecovery-statefulset.yaml
```diff
@@ -139,7 +139,7 @@ spec:
         - >
           bin/apply-config-from-env.py conf/bookkeeper.conf;
           {{- include "pulsar.autorecovery.zookeeper.tls.settings" . | nindent 10 }}
-          exec bin/bookkeeper autorecovery
+          OPTS="${OPTS} -Dlog4j2.formatMsgNoLookups=true" exec bin/bookkeeper autorecovery
```
michaeljmarshall, line 142: I was going to suggest that we update an argument elsewhere in a configmap, but by changing it this way, we guarantee that all users get this JVM arg.

michaeljmarshall (approved):
LGTM.

merlimat (approved):

PR #195, opened 2022-01-04
[CI] Improve logging in CI scripts and add timeouts

lhotari:
### Motivation

Currently it's hard to investigate CI failures since there isn't sufficient logging.

### Modifications

- show Kubernetes events every 15 seconds
- dump logs for all pods in the pulsar namespace every 5 minutes
- dump logs for all pods in the pulsar namespace when timeout occurs
- since CI scripts might be run locally, add support for "timeout" on MacOSX via instructing the user to install the coreutils brew package


eolivelli (approved):
Very useful 

Thanks

PR #196, opened 2022-01-05
[CI] Remove --ignore-errors=true parameter which requires newer kubectl

lhotari:
- #192 changes with upgraded kubectl version would be required unless
  the parameter is removed


lhotari:
It turns out that the current version of kubectl doesn't support `--prefix=true` which adds the pod name and container name as a prefix to each log line. It's better to get #192 merged to fix the kubectl log dumping solution which was added by #195

PR #203, opened 2022-01-17
Change default podManagementPolicy to Parallel for Zookeeper

lhotari:
### Motivation

- For speeding up startup, podManagementPolicy should be set to `Parallel` for Zookeeper

### Modifications

- set `zookeeper.podManagementPolicy` to `Parallel` in the default `values.yaml` file.

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -292,7 +292,7 @@ zookeeper:
   replicaCount: 3
   updateStrategy:
     type: RollingUpdate
-  podManagementPolicy: OrderedReady
+  podManagementPolicy: Parallel
   # If using Prometheus-Operator enable this PodMonitor to discover zookeeper scrape targets
   # Prometheus-Operator does not add scrape targets based on k8s annotations
   podMonitor:


eolivelli (approved):

codelipenghui (approved):

rmyates:
Note for future googlers, this change will break a helm chart upgrade if you installed the chart before this change due to podManagementPolicy in a StatefulSet being an immutable field.

> Error: UPGRADE FAILED: cannot patch "pulsar-zookeeper" with kind StatefulSet: StatefulSet.apps "pulsar-zookeeper" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden

Set it back to the previous version with the below in your custom values file which I don't think does any harm for existing installs

```
zookeeper:
  # Changed to Parallel by default in https://github.com/apache/pulsar-helm-chart/pull/203
  # However since we installed the chart prior to this and this field is immutable
  # set it back to the old value.
  podManagementPolicy: OrderedReady
```

michaeljmarshall:
@rmyates - thank you for posting this message. This change was intended to be backwards compatible, so it is valuable feedback to know it wasn't. Now that we've changed the default, I don't believe we can go back without also breaking new deployments that inherited the `Parallel` default.

rmyates:
@michaeljmarshall No worries. I think keeping the change for new users/installs is fine and does make sense looking at the [k8s docs](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#parallel-pod-management).

It's just one of those things where an immutable field bites you every now and then! I suspect they aired on the side of caution with changes to existing StatefulSets given the potential for destruction.

PR #213, opened 2022-01-26
Don't enable pulsar manager by default

lhotari:
### Motivation

- because of security reasons
  - it increases the attack surface
  - pulsar manager cannot be run as non-root user, issue reported as https://github.com/apache/pulsar-manager/issues/407
- it's an unnecessary feature for most users
  - wasted resource consumption


eolivelli (approved):
+1

michaeljmarshall:
+1

PR #215, opened 2022-01-26
[CI] Collect and upload k8s logs on failure

lhotari:
- collect and upload k8s logs when the GitHub Actions CI job fails or gets cancelled


michaeljmarshall (approved):
LGTM

PR #216, opened 2022-01-26
[CI] Fix issue with k8s log collection

lhotari:
- slash needs to be replaced with underscore
- fixes small issue in #215 solution


lhotari:
fixes this issue: https://github.com/apache/pulsar-helm-chart/runs/4955444589?check_suite_focus=true#step:7:6

michaeljmarshall (approved):

PR #220, opened 2022-01-31
Increase Zookeeper probe timeouts

lhotari:
### Motivation

- 5 seconds seems to be too short a probe timeout on a system with low resources, such as in CI

### Modifications

- change default timeout to 30 seconds (which matches the probe's periodSeconds parameter)

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -320,19 +320,19 @@ zookeeper:
       failureThreshold: 10
       initialDelaySeconds: 20
       periodSeconds: 30
-      timeoutSeconds: 5
+      timeoutSeconds: 30
     readiness:
       enabled: true
       failureThreshold: 10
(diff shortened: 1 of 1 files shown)


lhotari:
> > 5 seconds seems to be too short a probe timeout on a system with low resources, such as in CI
> 
> Were you seeing CI errors because of the low value? What are the consequences of changing this default? It seems to me that it'd let an unresponsive ZK receive traffic for 25 seconds longer, but zk might have some other mechanism that intercedes first. If it does increase the time to recovery, I am not sure that we should change the default for the chart. Can we update a CI `values.yaml` instead?

There hasn't been a timeout for k8s <1.20 before #214 . It's better to set a safe default than have a very tight timeout.

michaeljmarshall (approved):
@lhotari - thanks for the explanation, LGTM

PR #221, opened 2022-01-31
[Proxy] Configure metadata settings for Proxy

lhotari:
### Motivation

- enables service discovery in the proxy
- required by https://github.com/apache/pulsar/pull/14078 changes

### Modifications

- configure `zookeeperServers` and `configurationStoreServers` for the Proxy to enable service discovery

### DIFF
--- a/charts/pulsar/templates/proxy-configmap.yaml
+++ b/charts/pulsar/templates/proxy-configmap.yaml
@@ -31,8 +31,7 @@ data:
   zookeeperServers: "{{ template "pulsar.zookeeper.connect" . }}{{ .Values.metadataPrefix }}"
   {{- if .Values.pulsar_metadata.configurationStore }}
   configurationStoreServers: "{{ template "pulsar.configurationStore.connect" . }}{{ .Values.pulsar_metadata.configurationStoreMetadataPrefix }}"
-  {{- end }}
-  {{- if not .Values.pulsar_metadata.configurationStore }}
+  {{- else }}
   configurationStoreServers: "{{ template "pulsar.zookeeper.connect" . }}{{ .Values.metadataPrefix }}"
   {{- end }}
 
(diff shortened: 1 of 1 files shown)


charts/pulsar/templates/proxy-configmap.yaml
```diff
@@ -27,6 +27,15 @@ metadata:
     {{- include "pulsar.standardLabels" . | nindent 4 }}
     component: {{ .Values.proxy.component }}
 data:
+  # Metadata settings
+  zookeeperServers: "{{ template "pulsar.zookeeper.connect" . }}{{ .Values.metadataPrefix }}"
+  {{- if .Values.pulsar_metadata.configurationStore }}
+  configurationStoreServers: "{{ template "pulsar.configurationStore.connect" . }}{{ .Values.pulsar_metadata.configurationStoreMetadataPrefix }}"
+  {{- end }}
+  {{- if not .Values.pulsar_metadata.configurationStore }}
```
michaeljmarshall, line 35: Nit:
```suggestion
  {{- else }}
```

michaeljmarshall (approved):

charts/pulsar/templates/proxy-configmap.yaml
lhotari: Thanks, that makes sense. I had copy-pasted the original logic from https://github.com/apache/pulsar-helm-chart/blob/dc97bd4ac6ae59610f685341dedd716ed10a8165/charts/pulsar/templates/broker-configmap.yaml#L30-L37

michaeljmarshall (changes requested):
@lhotari - do you plan to complete this work? Since it is not generally recommended to configure the zookeeper connection, I don't think we should enable it by default. We could make it easier to configure, though.

PR #224, opened 2022-02-02
Allow us to turn on PodSecurityPolicy for Pulsar clusters with the same name deployed in different K8S namespaces in the same cluster

frankjkelly:
(PodSecurityPolicy is a cluster-level-resource so we are creating new PSPs by appending the pulsar namespace)

Fixes #<xyz>

### Motivation

* We want to be able to deploy multiple Pulsar clusters in different K8S namespaces but having the same helm release name. 

### Modifications

* PodSecurityPolicies are cluster-wide so to avoid clashing we need to distinguish each Pulsar Cluster PSP somehow so we appended the `pulsar.namespace` value
* For backwards compatiblity we keyed off of `rbac.limit_to_namespace` which by default is `false`
https://github.com/apache/pulsar-helm-chart/blob/master/charts/pulsar/values.yaml#L93
### Verifying this change

### DIFF
--- a/charts/pulsar/Chart.yaml
+++ b/charts/pulsar/Chart.yaml
@@ -21,7 +21,7 @@ apiVersion: v2
 appVersion: "2.7.4"
 description: Apache Pulsar Helm chart for Kubernetes
 name: pulsar
-version: 2.7.10
+version: 2.7.11
 home: https://pulsar.apache.org
 sources:
 - https://github.com/apache/pulsar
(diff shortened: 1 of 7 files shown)


sijie (approved):

ckdarby:
Just an FYI, from [the docs](https://kubernetes.io/docs/concepts/policy/pod-security-policy/):
> PodSecurityPolicy is deprecated as of Kubernetes v1.21, and will be removed in v1.25.

frankjkelly:
@ckdarby good point `1.25` will be coming at some point later this year `1.24` looks due in April https://github.com/kubernetes/sig-release/tree/master/releases/release-1.24
and with a release cycle every 15 weeks give or take that puts us I think in July/August timeframe if my math is correct.
https://kubernetes.io/blog/2021/07/20/new-kubernetes-release-cadence/

lhotari (approved):

PR #225, opened 2022-02-07
allow specifying the nodeSelector for the init jobs

elangelo:
### Motivation

When deploying pulsar to an AKS cluster with windows nodepools i was unable to specify that the Jobs of the initalize release had to run on linux nodes. With the change i can now specify a node selector for the init jobs.

### Modifications

add nodeSelector on pulsar_init and bookie_init

### Verifying this change

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -652,6 +652,9 @@ pulsar_metadata:
   # configurationStore:
   configurationStoreMetadataPrefix: ""
   configurationStorePort: 2181
+  ## optional you can specify a nodeSelector for all init jobs (pulsar-init & bookkeeper-init)
+  # nodeSelector:
+    # cloud.google.com/gke-nodepool: default-pool
 
   ## optional, you can provide your own zookeeper metadata store for other components
   # to use this, you should explicit set components.zookeeper to false
--- a/charts/pulsar/templates/bookkeeper-cluster-initialize.yaml
+++ b/charts/pulsar/templates/bookkeeper-cluster-initialize.yaml
@@ -32,6 +32,9 @@ spec:
     {{- if and .Values.rbac.enabled .Values.rbac.psp }}
       serviceAccountName: "{{ template "pulsar.fullname" . }}-{{ .Values.bookkeeper.component }}"
     {{- end }}
+      {{- if .Values.initialize_jobs.nodeSelector }}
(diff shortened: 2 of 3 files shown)


lhotari (changes requested):
Thanks for the PR. Would it be possible to use the key `pulsar_metadata.nodeSelector` instead of `initialize_jobs.nodeSelector`? Please also add a placeholder in the default values.yaml with a comment documenting the setting. The value should be empty.

elangelo:
> Thanks for the PR. Would it be possible to use the key `pulsar_metadata.nodeSelector` instead of `initialize_jobs.nodeSelector`? Please also add a placeholder in the default values.yaml with a comment documenting the setting. The value should be empty.

what do i do for the bookkeeper_cluster_initalize jobs? specify another nodeSelector value? or use the pulsar_metadata.nodeSelector value there as well?

lhotari:
> what do i do for the bookkeeper_cluster_initalize jobs? specify another nodeSelector value? or use the pulsar_metadata.nodeSelector value there as well?

I think it's fine to use the pulsar_metadata.nodeSelector value there as well.

lhotari (approved):
LGTM, good work @elangelo

charts/pulsar/values.yaml
```diff
@@ -652,6 +652,9 @@ pulsar_metadata:
   # configurationStore:
   configurationStoreMetadataPrefix: ""
   configurationStorePort: 2181
+  ## optional you can specify a nodeSelector for all init jobs (pulsar-init & bookkeeper-init)
+  # nodeSelector:
+    # cloud.google.com/gke-nodepool: default-pool
```
michaeljmarshall, line 657: Nit: I think it is valuable to add the key to the values file.

```suggestion
  nodeSelector: {}
```

michaeljmarshall (commented):
LGTM. Left a minor comment. Thanks for your contribution @elangelo!

charts/pulsar/values.yaml
michaeljmarshall: This is a larger style issue that can be addressed later. Going to merge this PR without worrying about this fix.

michaeljmarshall (approved):
LGTM

michaeljmarshall:
Thanks for your contribution @elangelo!

PR #204, opened 2022-01-17
Upgrade cert-manager to 1.5.4

lhotari:
### Motivation

- Keep cert-manager version updated

### Modifications

- upgrade cert-manager to 1.5.4 version
- use apiVersion `cert-manager.io/v1`

PR #233, opened 2022-02-17
feat(certs): use actual v1 spec for certs

smazurov:
### Motivation

In #204, api version of the cert resources was updated to v1. This was insufficient because `v1` has different spec from `v1alpha1` 

This MR finishes the work that #204 and @lhotari started.

### Modifications

Changed the spec of certs to match v1 cert manager spec.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/tls-certs-internal.yaml
+++ b/charts/pulsar/templates/tls-certs-internal.yaml
@@ -31,15 +31,17 @@ spec:
   secretName: "{{ .Release.Name }}-{{ .Values.tls.proxy.cert_name }}"
   duration: "{{ .Values.tls.common.duration }}"
   renewBefore: "{{ .Values.tls.common.renewBefore }}"
-  organization:
-{{ toYaml .Values.tls.common.organization | indent 2 }}
+  subject:
+    organizations:
+{{ toYaml .Values.tls.common.organization | indent 4 }}
   # The use of the common name field has been deprecated since 2000 and is
   # discouraged from being used.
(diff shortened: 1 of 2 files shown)


lhotari:
@smazurov Thanks for the contribution. I created a PR to first revert #204 changes so that we have time to prepare a better solution for migrating to use `cert-manager.io/v1` api version in a backwards compatible way.

It seems that the backwards compatible migration might require more changes. For new installations (`{{- if .Release.IsInstall }}`) it should be fine to use the `cert-manager.io/v1` api version. For existing installations, I'm not exactly sure to handle the backwards compatibility. 
cert-manager documentation about the deprecated api versions: https://cert-manager.io/docs/installation/upgrading/remove-deprecated-apis/#upgrading-existing-cert-manager-resources

smazurov:
assuming kube is aware of both apis, upgrading is straightfoward. So the change is fine as long as the new spec for v1 is actually followed. Once cert-manager is upgraded to v1.6+, they dropped support for alpha versions, so it won't let you upgrade.

youg9203:
Is the PR still in progress? Any estimation of when it will be released? I have been facing issues with TLS and if I'm going to use the old version `v1.1.0` then it's not working with the new pulsar `2.9.0`.  Any idea which config will work with the newer pulsar version?

toneill818 (approved):

michaeljmarshall (approved):
LGTM. In my view, this change is backwards compatible since the net effect is just an updated k8s secret for the TLS certs.

michaeljmarshall:
I am going to merge this now, and then I'll test it in https://github.com/apache/pulsar-helm-chart/pull/306

PR #238, opened 2022-02-21
Cert Fix

toneill818:
### Motivation

In the scripts folder there is a script to install [cert-manager 1.5.4](https://cert-manager.io/v1.5-docs/usage/certificate/), but the chart's templates do not work with that version.

### Modifications

- Moved `keyAlgorithm`, `keySize`, and `keyEncoding` under the `privateKey` object
- Moved `organization` under subject and renamed to `organizations`
- Bumped Chart Version

### Verifying this change

- [X] Make sure that the change passes the CI checks.

PR #239, opened 2022-02-28
Revert "Upgrade cert-manager to 1.5.4 (#204)"

lhotari:
This reverts commit 40a4d50923bd93c960a4ef940930177ca80cb77e (#204).

Supporting `cert-manager.io/v1` api version requires changes made in #233 or #238. In addition, the change should be made in a way that it is backwards compatible.

### DIFF
--- a/charts/pulsar/Chart.yaml
+++ b/charts/pulsar/Chart.yaml
@@ -21,7 +21,7 @@ apiVersion: v2
 appVersion: "2.7.4"
 description: Apache Pulsar Helm chart for Kubernetes
 name: pulsar
-version: 2.7.13
+version: 2.7.14
 home: https://pulsar.apache.org
 sources:
 - https://github.com/apache/pulsar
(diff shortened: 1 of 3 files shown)


toneill818:
@lhotari Would you be interested in having cert-manager installed as a dependency if enabled, or kept outside this chart? If we require it as a dependency we can use the following to get the version (we will need to loop through in case more dependencies are added in the future).
```
{{- $certManger := index .Chart.Dependencies 0 -}}
{{ $certManger.Version }}
```
Then we could use the [semver](http://masterminds.github.io/sprig/semver.html) function to get more information about the release and edit the TLS to be backwards compatible. 
`{{- if semverCompare "^1.5.0"  $certManger.Version -}}`

michaeljmarshall:
I merged https://github.com/apache/pulsar-helm-chart/pull/233, so we can close this.

PR #242, opened 2022-03-10
Fix issue that bookkeeper don't support some jvm options anymore on java11

kwenzh:
### Motivation
in pulsar 2.8.x,  there are a few JVM options that the bookkeeper doesn't support anymore. It should be removed from the vaules.yaml as the default configuration. 

suggest helm template default values.yaml keep same change. 

fix in : https://github.com/streamnative/charts/pull/403

### Modifications

in bookie jcm config, remove jvm params:
```
-XX:+PrintGCDetails
-XX:+PrintGCTimeStamps
-XX:+PrintGCApplicationStoppedTime
-XX:+PrintHeapAtGC
 -Xloggc:/var/log/bookie-gc.log
 -XX:G1LogLevel=finest
```
and add jvm params
```
 -Xlog:gc*
-Xlog:::time,level,tags
-Xlog:safepoint
-Xlog:gc+heap=trace
-Xlog:gc:/var/log/bookie-gc.log
```

### Verifying this change

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -567,10 +567,10 @@ bookkeeper:
       -XX:-ResizePLAB
       -XX:+ExitOnOutOfMemoryError
       -XX:+PerfDisableSharedMem
-      -XX:+PrintGCDetails
-      -XX:+PrintGCTimeStamps
-      -XX:+PrintGCApplicationStoppedTime
-      -XX:+PrintHeapAtGC
+      -Xlog:gc*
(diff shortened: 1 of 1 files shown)


michaeljmarshall:
Thanks for your contribution @kwenzh! It appears the change was already contributed here https://github.com/apache/pulsar-helm-chart/pull/255. Sorry for the delayed review. I am going to close this since the desired changes are already merged. Please re-open if necessary. Thanks!

Issue #229, opened 2022-02-16
Update pulsar-manager-service template to allow for admins to open multiple ports


bsheltonihs:
**Is your feature request related to a problem? Please describe.**
I am using argo-cd to deploy on openshift using these helm charts. The issue I have is the backend admin service port is `7750` which the `pulsar-manager-service.yaml` template doesn't expose. I am trying to follow the instructions here: [set-administrator-account-and-password](https://pulsar.apache.org/docs/en/administration-pulsar-manager/#set-administrator-account-and-password) from a node in the cluster but it doesn't work. When I go into openshift and manually add the `7750` port to the service I can then set the superuser info.  

**Describe the solution you'd like**
I would like the `pulsar-manager-service.yaml` file to be updated from:
 `ports:
    - name: server
      port: {{ .Values.pulsar_manager.service.port }}
      targetPort: {{ .Values.pulsar_manager.service.targetPort }}
      protocol: TCP`
to:
`ports:
    - name: server
      port: {{ .Values.pulsar_manager.service.port }}
      targetPort: {{ .Values.pulsar_manager.service.targetPort }}
      protocol: TCP
    {{- if .Values.pulsar_manager.service.admin_port }}
    - name: {{ .Values.pulsar_manager.service.admin_port.name }}
      port: {{ .Values.pulsar_manager.service.admin_port.port }}
      targetPort: {{ .Values.pulsar_manager.service.admin_port.targetPort }}
      protocol: TCP
    {{- end }}`

**Describe alternatives you've considered**
The only other options I have are:  
1. Create and maintain my own helm charts (which long term isn't viable)
2. Do this manually (which doesn't allow it to be automated like I need)

**Additional context**
Even if you want to hardcode it instead of making it virialized that would be fine also.

alpreu:
Unfortunately, pulsar-manager does not really support OpenShift yet. You can try @streamvisor, it runs on OpenShift out of the box and is also free :)

lhotari:
> Unfortunately, pulsar-manager does not really support OpenShift yet. You can try @streamvisor, it runs on OpenShift out of the box and is also free :)

Please stop spamming this issue tracker with your advertisements, @alpreu.

lhotari:
Closing this outdated issue. #457 improves the pulsar manager integration.

PR #249, opened 2022-03-24
Expose pulsar manager backend port

larmic:
Fixes #<xyz>

### Motivation

To create a pulsar manager admin user, port 7750 must be available. 

See also https://github.com/apache/pulsar-helm-chart/issues/229.

### Modifications

I split the existing pulsar-manager-service.yaml into frontend and backend and made the backend configurable via a config parameter (pulsar_manager.serviceAdmin.enabled).

### Verifying this change


charts/pulsar/templates/pulsar-manager-service.yaml
```diff
@@ -21,7 +21,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: "{{ template "pulsar.fullname" . }}-{{ .Values.pulsar_manager.component }}"
+  name: "{{ template "pulsar.fullname" . }}-{{ .Values.pulsar_manager.component }}-frontend"
```
michaeljmarshall, line 24: Changing this service name might break existing integrations. Can we leave it as is in this PR?

michaeljmarshall (commented):
Thank you for your contribution @larmic! Sorry for the delayed review. Adding the service looks good to me, but I am not sure that we should change the existing service's name since it would be a breaking change.

keenborder786:
Is their any update on this issue? This is still a problem in the main helm chart.

charts/pulsar/templates/pulsar-manager-service.yaml
svabra: @michaeljmarshall Hi, Michael, I fail to understand why this is a breaking change. It's an extension, not a change (apropos open-close-principle). If there is a reason why this would break, what else do you suggest we could submit in the next PR? Any strategy from the team?

lhotari:
Closing this as outdated. #457 fixes the issue.

PR #258, opened 2022-04-12
Fix PodMonitor name conflicts for multiple releases in same namespace

### LINKED ISSUE #257: podMonitor name conflicts for multiple releases in the same namespace
**Describe the bug**
Install two releases of pulsar cluster with podMonitor enabled in the same namespace will fail.

**To Reproduce**
1. install release `test-1` in default namespace
`helm install test-1 apache/pulsar --set broker.podMonitor.enabled=true`
this will success
2. install release `test-2` in default namespace
`helm install test-2 apache/pulsar --set broker.podMonitor.enabled=true`
This will fail due to the following error:
```
Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists. Unable to continue with install: PodMonitor "pulsar-broker" in namespace "default" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "test-2": current value is "test-1"
```

**Expected behavior**
Expect the second installment will also be successful.

**Screenshots**

apache/pulsar chart version: 2.9.2

**Additional context**

LoveEachDay:
Signed-off-by: Edward Zeng <jie.zeng@zilliz.com>

Fixes #257

### Motivation

Fix PodMonitor name conflicts for multiple releases in same namespace

### Modifications

Use release name instead of hardcode `pulsar.name` for pod monitor name.

### Verifying this change


michaeljmarshall (approved):
LGTM. Thanks for your contribution @LoveEachDay!

I believe this will this change the value of the `job` label in prometheus. I don't see that as a problem, but it is something we'll want to mention in the release notes.

charts/pulsar/templates/bookkeeper-podmonitor.yaml
```diff
@@ -22,7 +22,7 @@
 apiVersion: monitoring.coreos.com/v1
 kind: PodMonitor
 metadata:
-  name: {{ template "pulsar.name" . }}-bookie
+  name: {{ .Release.Name }}-bookie
```
michaeljmarshall, line 25:
```suggestion
  name: {{ template "pulsar.fullname" . }}-bookie
```

michaeljmarshall (commented):
@LoveEachDay - sorry for the delay here. I think we should instead make this use `{{ template "pulsar.fullname" . }}` for the prefix, since that is what each of the components themselves use. It will use the `.Release.Name` in certain cases. I am getting ready to work on releasing a new version of the helm chart, so if you're able to update this PR soon, we should be able to get this released. Thanks!

michaeljmarshall:
In order to include this in the upcoming release, I committed my suggestion.

michaeljmarshall:
We failed the lint-test for the previous commit due to an unexpected error. I merged master into this branch to get the latest tests to run.

michaeljmarshall:
Closing and re-opening to get the latest changes to the CI pipeline.

PR #251, opened 2022-03-24
make proxy httpNumThreads configurable

filipecaixeta:
Fixes https://github.com/apache/pulsar-helm-chart/issues/250

### Motivation

`httpNumThreads` is hardcoded to 8 in `charts/pulsar/templates/proxy-configmap.yaml`
When trying to override in `values.yaml` by using `proxy.configData.httpNumThreads` we get an error because the keys get duplicated.
This happens because `{{ toYaml .Values.proxy.configData | indent 2 }}` doesn't deduplicate the keys and there is no other way to set `httpNumThreads`

### Modifications

Removing the key from charts/pulsar/templates/proxy-configmap.yaml and adding it to the values yaml solves the problem.

### Verifying this change

- [x] Make sure that the change passes the CI checks.

PR #259, opened 2022-04-14
[Proxy] Increase httpNumThreads from 8 to 16

lhotari:
- after the recent changes in Pulsar Proxy, the proxy won't start if httpNumThreads is set to 8
  and there are >= cores assigned to the pod

Error message is 
```
Caused by: java.lang.IllegalStateException: Insufficient configured threads: required=8 < max=8 for WebExecutorThreadPool[etp884604029]@34b9fc7d{STARTED,8<=8<=8,i=2,q=0,ReservedThreadExecutor@3fcee3d9{s=0/1,p=0}}
```

### DIFF
--- a/charts/pulsar/templates/proxy-configmap.yaml
+++ b/charts/pulsar/templates/proxy-configmap.yaml
@@ -28,7 +28,7 @@ metadata:
     component: {{ .Values.proxy.component }}
 data:
   clusterName: {{ template "pulsar.cluster.name" . }}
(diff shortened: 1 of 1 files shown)


michaeljmarshall (commented):
Would it make more sense to make it configurable?

michaeljmarshall:
Superseded by https://github.com/apache/pulsar-helm-chart/pull/251, which makes the value configurable.

PR #270, opened 2022-06-14
Enable the pulsar manager in the minikube values

zymap:
---

Fixes: https://github.com/apache/pulsar/issues/15927

### Motivation

We have documented the using pulsar manager in the Getting started
with helm on the pulsar website. We should enable the pulsar manager
by default in the minikube values.

### Modifications

- enable the pulsar manager by default in the minikube values.

### DIFF
--- a/examples/values-minikube.yaml
+++ b/examples/values-minikube.yaml
@@ -28,6 +28,7 @@ affinity:
 # disable auto recovery
 components:
   autorecovery: false
+  pulsar_manager: true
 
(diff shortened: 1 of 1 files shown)


michaeljmarshall (approved):
LGTM. It looks like the version is still 0.2.0. Does it need to be increased?

zymap:
Maybe we should upgrade the pulsar-manager to 0.3.0. I can upgrade it in another PR

PR #274, opened 2022-06-28
Bump Apache Pulsar 2.10.1

codelipenghui:


Anmol057:
Hi, when can we expect 2.10.1 to be released?

dan-autonomic:
Hi Team,
Just bumping this since 2.10.1 has really good features, like TableView not available to previous versions.

charts/pulsar/Chart.yaml
```diff
@@ -18,10 +18,10 @@
 #
 
 apiVersion: v2
-appVersion: "2.9.2"
+appVersion: "2.10.1"
 description: Apache Pulsar Helm chart for Kubernetes
 name: pulsar
-version: 2.9.3
+version: 2.10.1
```
michaeljmarshall, line 24: Can we decouple this PR from the version bump? I know we have historically had a process to bump the version here and trigger a release, but I think we should make releases align with the official Apache Release process.

michaeljmarshall (commented):
@codelipenghui - did you experience any issues while testing the upgrade path? Is there any reason we cannot merge this PR? Thanks!

charts/pulsar/Chart.yaml
michaeljmarshall: Since I am working on the new release process, I added a commit to remove this change.

michaeljmarshall:
@Anmol057 @dan-autonomic - I cannot make any guarantees about when this chart will be released, but I am hoping to get a release available within a couple weeks. Note that you can likely use Apache Pulsar 2.10.1 with the currently available helm chart. It is always a good idea to have your own tests, and if you are upgrading an existing cluster, it is a good idea to verify on a non-production cluster that your in place upgrades work as expected for your use cases.

michaeljmarshall:
@lhotari - looks like the TLS tests are failing for 2.10.1, but not for 2.9.3. I see that https://github.com/apache/pulsar-helm-chart/pull/223 is related, but I haven't figured out the root issue yet. The core error is coming from the liveness probe updated in that PR:

```
Liveness probe failed: 140472017823040:error:0200206F:system library:connect:Connection refused:../crypto/bio/b_sock2.c:110:
140472017823040:error:2008A067:BIO routines:BIO_connect:connect error:../crypto/bio/b_sock2.c:111:
140472017823040:error:0200206F:system library:connect:Connection refused:../crypto/bio/b_sock2.c:110:
140472017823040:error:2008A067:BIO routines:BIO_connect:connect error:../crypto/bio/b_sock2.c:111:
connect:errno=111
```

I plan to continue digging, but I wanted to share in case you have seen this already.

michaeljmarshall:
The test might just be flaky. I added some extra logging here https://github.com/apache/pulsar-helm-chart/pull/304 (I meant to make the PR in my fork, but it's in the official project), and the zookeeper started without any obvious issues: https://github.com/apache/pulsar-helm-chart/actions/runs/3270514005/jobs/5379215371. The test still failed though.

michaeljmarshall (approved):

PR #280, opened 2022-07-20
Make included prometheus server configurable

### LINKED ISSUE #278: Make included Prometheus server configurable
**The prometheus configuration is currently hardcoded. Extending it is not possible - e.g. connecting it with an external alert manager.**
We want to connect the included Prometheus server to an external Alert Manager which is currently not possible.

Allow to modify the prometheus configuration entirely in `values.yaml`

We think that making only external Alert Managers configurable would not catch all possible wishes against prometheus configuration.

**Additional context**
-

MBcom:
Fixes #278

### Motivation

We want to connect the included Prometheus server to an external Alert Manager which is currently not possible.

### Modifications

We have moved the hard coded configuration from `templates/prometheus-configmap.yaml` to `prometheus.configData` in `values.yaml`

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/prometheus-configmap.yaml
+++ b/charts/pulsar/templates/prometheus-configmap.yaml
@@ -29,38 +29,5 @@ metadata:
 data:
   # Include prometheus configuration file, setup to monitor all the
   # Kubernetes pods with the "scrape=true" annotation.
-  prometheus.yml: |
-    global:
-      scrape_interval: 15s
-    scrape_configs:
-    - job_name: 'prometheus'
-      static_configs:
-      - targets: ['localhost:9090']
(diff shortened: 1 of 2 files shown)


michaeljmarshall:
Thank you for your contribution @MBcom! I am proposing that we use the `kube-prometheus-stack` to manage prometheus and grafana. I think this will fulfill the needs raised in this PR and in #278. If you agree, I think we can close this PR.

michaeljmarshall:
Since we merged https://github.com/apache/pulsar-helm-chart/issues/278, the Prometheus server's configuration is now exposed, so I am going to close this PR. Please re-open it if you believe this is incorrect @MBcom. Thanks for your contribution!

PR #286, opened 2022-08-30
Allow to use selectors with volumeClaimTemplates

claudio-vellage:
### Motivation

Currently it's not possible to use selectors with volumeClaimTemplates which makes it hard/impossible to bind statically provisioned PVs.

### Modifications

Added (optional) selectors to `volumeClaimTemplates` and documented in values file.

### Verifying this change


claudio-vellage:
@lhotari is there anything left to do for me? Can this be assigned to someone for review? It's a rather small change which doesn't affect existing deployments, only adds additional functionality. It allows that selectors can be added to volumeClaimTemplates, which will allow to easily create PVCs which bind to existing PVs using selectors. This allows to reuse the same PVs even if all pulsar resources are deleted. This will be more robust for production deployments and avoid having to restore the content of the PVs manually in case of disaster recovery.

charts/pulsar/values.yaml
```diff
@@ -383,6 +383,10 @@ zookeeper:
         # type: pd-ssd
         # fsType: xfs
         # provisioner: kubernetes.io/gce-pd
+      ## If you want to bind static persistent volumes via selectors, e.g.:
+      # selectors:
+        # matchLabels:
+        # app: pulsar-zookeeper
```
michaeljmarshall, line 389: I think it is slightly more helpful to have a good zero value here. Also, we need `selector` not `selectors`.

```suggestion
      ## If you want to bind static persistent volumes via selectors, e.g.:
      selector: {}
```

charts/pulsar/values.yaml
```diff
@@ -515,6 +519,10 @@ bookkeeper:
         # type: pd-ssd
         # fsType: xfs
         # provisioner: kubernetes.io/gce-pd
+      ## If you want to bind static persistent volumes via selectors, e.g.:
+      # selectors:
+        # matchLabels:
+        # app: pulsar-bookkeeper-journal
```
michaeljmarshall, line 525:
```suggestion
      selector: {}
```

michaeljmarshall (changes requested):
@claudio-vellage - thank you for your contribution. Sorry for the delayed review here. Once we get the `values.yaml` cleaned up, I plan to merge this. Thanks.

charts/pulsar/values.yaml
claudio-vellage: I've made the changes, please review

charts/pulsar/values.yaml
claudio-vellage: I've made the changes, please review

michaeljmarshall (approved):
LGTM

michaeljmarshall:
@claudio-vellage - are you able to resolve the merge conflicts? Thanks!

michaeljmarshall:
Closing and reopening to get the latest test changes. (I am pretty sure we don't need to merge master into this branch to get the fix I just merge here #315, but that might be necessary.)

PR #291, opened 2022-09-26
Alphabetically sort list of super users

### LINKED ISSUE #288: superUserRoles order is unpredictable
**Describe the bug**
When specifying multiple roles in `.Values.auth.superUsers` the values are converted to a comma-separated list by piping the dict through `values` and `join` in helm templating, `values` however doesn't guarantee that the order of elements will be the same every time. Therefor it recommends also passing it through `sortAlpha` to sort the list alphabetically.

This is a problematic when `.Values.broker.restartPodsOnConfigMapChange` is enabled because the checksum of the configmap changes every time the list's order is changed, resulting in the statefulsets rolling out a new version of the pods.

Relevant code:
https://github.com/apache/pulsar-helm-chart/blob/c059ea25d80d6165b67ee163252a682198d8933f/charts/pulsar/templates/broker-configmap.yaml#L113
https://github.com/apache/pulsar-helm-chart/blob/c059ea25d80d6165b67ee163252a682198d8933f/charts/pulsar/templates/proxy-configmap.yaml#L68

**To Reproduce**
1. Use following values:
  ```yaml
  broker:
    restartPodsOnConfigMapChange: true
  auth:
    authentication:
      enabled: true
    superUsers:
      broker: "broker-admin"
      proxy: "proxy-admin"
      client: "admin"
  ```
2. `helm template` a few times
3. Notice the `checksum/config` annotation in broker statefulset change often without any change to values.yaml.

**Expected behavior**
No changes to the templated output to happen when no values changes happen.

**Additional context**
We use ArgoCD to automatically sync our pulsar application every 10 minutes which is causing the pods to restart almost every time.

arnarg:
Fixes #288 

### Motivation

When specifying multiple roles in `.Values.auth.superUsers` the values are converted to a comma-separated list by piping the dict through `values` and `join` in helm templating, `values` however doesn't guarantee that the order of elements will be the same every time. Therefor it recommends also passing it through `sortAlpha` to sort the list alphabetically.

This is a problematic when `.Values.broker.restartPodsOnConfigMapChange` is enabled because the checksum of the configmap changes every time the list's order is changed, resulting in the statefulsets rolling out a new version of the pods.

### Modifications

Pass list through `sortAlpha`.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/broker-configmap.yaml
+++ b/charts/pulsar/templates/broker-configmap.yaml
@@ -110,7 +110,7 @@ data:
   authenticationEnabled: "true"
   {{- if .Values.auth.authorization.enabled }}
   authorizationEnabled: "true"
-  superUserRoles: {{ .Values.auth.superUsers | values | join "," }}
+  superUserRoles: {{ .Values.auth.superUsers | values | sortAlpha | join "," }}
   {{- end }}
   {{- if eq .Values.auth.authentication.provider "jwt" }}
   # token authentication configuration
--- a/charts/pulsar/templates/proxy-configmap.yaml
+++ b/charts/pulsar/templates/proxy-configmap.yaml
@@ -64,7 +64,7 @@ data:
   # disable authorization on proxy and forward authorization credentials to broker
   authorizationEnabled: "false"
   forwardAuthorizationCredentials: "true"
-  superUserRoles: {{ .Values.auth.superUsers | values | join "," }}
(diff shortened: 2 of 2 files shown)


michaeljmarshall (approved):
LGTM. Thank you for your contribution @arnarg!

arnarg:
Doesn't seem like the failing workflows are because of my changes, right? @michaeljmarshall

michaeljmarshall:
No, these might just be flaky workflows. I triggered the failed jobs to run again.

Issue #290, opened 2022-09-22
Release Pulsar Helm Chart on dist.apache.org


tisonkun:
## Motivation

Currently, we host Pulsar Helm Chart's metadata (`index.yaml`) on https://pulsar.apache.org/charts/index.yaml. Users encountered an issue today (https://github.com/apache/pulsar/issues/17799) due to we removed the file during "clean up" content before building the official website fully.

Depending on another repo is brittle. It's hard to ensure further development happening in the pulsar-site repo is aware of this trick while this is critical to the pulsar-helm-chart usage.

## Proposal

I propose we adopt releasing Pulsar Helm Chart in GitHub Pages flavor to make this chart self-contained.

* Helm official document: https://helm.sh/docs/howto/chart_releaser_action/
* Chart release action: https://github.com/helm/chart-releaser-action
* A live example from OpenSearch:
    * https://github.com/opensearch-project/helm-charts/blob/main/.github/workflows/release.yaml
    * https://github.com/opensearch-project/helm-charts/blob/gh-pages/index.yaml
* Prototype on fork: https://github.com/tisonkun/pulsar-helm-chart

## Implementation

The implementation steps are:

1. Create `gh-pages` branch on pulsar-helm-chart repo, add current `index.yaml` file to the branch.
2. Ask Apache INFRA team to config the branch as the GitHub Pages branch.
3. Add the workflow using chart-releaser-action (bundled in this repo, it's lightweight).
4. Redirect https://pulsar.apache.org/charts to the new place.

## Risks

~~The last step _may_ be optional because IIUC helm chart's index is global. Thus once we publish the first new index with chart-releaser-action, all downloads to the chart will find the index from the new place. Although, I'm not an expert on Helm Chart so here requires some verification.~~ No. `https://pulsar.apache.org/charts` is actually documented at https://pulsar.apache.org/docs/helm-overview/#upgrading. So we should make the redirection.

Another risk is that the reliability of pulsar.apache.org is promised by the ASF INFRA team, while GitHub Pages are handled by the GitHub team. I'd say GitHub is less stable than ASF sites.

## Alternative

Perform an Apache release and upload the index to https://dist.apache.org/repos/dist/release/pulsar. This approach requires modifying homemade automation (dist.a.o uses SVN), and possibly brings new bugs.

tisonkun:
cc @lhotari @mattisonchao @yaalsn @ericsyh

michaeljmarshall:
[...]

PR #292, opened 2022-10-09
Replace handmade lint script with official action 

tisonkun:
This refers to #290.

The `helm/chart-releaser-action` update `index.yaml` in an incremental manner, so we need to push the existing `index.yaml` file to `gh-pages` later.


.github/workflows/release.yml
```diff
@@ -17,39 +17,36 @@
...
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
 
-      - name: Detect changed files
-        id:   changes
-        uses: apache/pulsar-test-infra/paths-filter@master
+      - name: Install Helm
+        uses: azure/setup-helm@v1
```
tisonkun, line 45: This action is too complex to bundle. An INFRA ticket already adds this action to the whitelist: https://issues.apache.org/jira/browse/INFRA-23716

.github/workflows/lint.yml
```diff
@@ -43,9 +42,22 @@ jobs:
...
+      - name: Set up Helm
+        if: ${{ steps.check_changes.outputs.docs_only != 'true' }}
+        uses: azure/setup-helm@v3
+        with:
+          version: v3.10.0
+
+      - name: Set up Python
         if: ${{ steps.check_changes.outputs.docs_only != 'true' }}
-        id: lint
-        uses: helm/chart-testing-action@v2.0.0
```
tisonkun, line 49: This action is never allowed by INFRA settings and thus we always failed before. Fix on the fly.

tisonkun:
cc @michaeljmarshall @ericsyh @mattisonchao @yaalsn

wolfstudy (approved):
Cool LGTM +1

tisonkun:
@michaeljmarshall Thanks for your follow-up. Let me check your investigation today and feedback. My first object is to decouple the Chart index file from the website :)

tisonkun:
> will break

Yep. As mentioned in #290, we can set a server-side redirection in the website `.htaccess`. The reason is that, website build process doesn't handle Chart yaml index so that it's possible we accidentally remove the file - they're totally different topics.

Setting up a redirection is in the domain of a website maintenance and avoid further development accidentally modifies it.

michaeljmarshall:
Sure, I agree that the current workflow is a bit of a hack, and the current `index.yaml` is not very well protected. I added a comment to #290, too.

tisonkun:
@michaeljmarshall I've narrow the scope to replace the lint scripts with `chart-testing-action`. PTAL.

I'm going through your proposal on release process now.

tisonkun:
Weird. Is it an unstable test?

tisonkun:
@michaeljmarshall I don't think the failures are related to this patch. Could you take a looks and rerun the tests?

michaeljmarshall:
@tisonkun - yes, the TLS Installation and ZK & BK TLS Only tests are unstable. I have spent some time looking into them this week, but I haven't found the root cause. It seems like there may be some instability with the environment because some of the errors are about refused connections to ports that appear to be working for some pods but not others.

michaeljmarshall:
Here is an issue for the flaky tests: #311

michaeljmarshall (approved):

Issue #294, opened 2022-10-13
[security] helm chart is outdated and includes images with 992 vulnerabilities (623 fixable)


hpvd:
**Describe the bug**
helm chart is outdated and includes images with 992 vulnerabilities

In detail:
- there are dependencies with well known security issues (with official CVE numbers)
- there is a pretty huge number of known and documented vulnerabilities: 992
-  including important ones (critical, high rating)
-  not only in the accompanying software in helm (prometheus, grafana) but in core directly (pulsar)
-  some were known for 9 years (CVE numbers from 2013)
- there are possible fixes for most of them (for 623)

see source https://artifacthub.io/packages/helm/apache/pulsar?modal=security-report

Of course, this is only a first rough impression given by this analysis.
And the chart does not contain the very latest version of pulsar (even so, it's the latest official helm chart)
When looking into every detail of the reported numbers, you can of course argue why not every counted vulnerabilities is a disaster...

=> But how can you easily argue to anyone having seen this fast result

>  In general, this software (pulsar) is secure, you can use it without any concerns.

?

This issue is a shortened copy from
Arguments for "why pulsar is secure?"
https://github.com/apache/pulsar/issues/18041
for more details and comments, please see there.

hpvd:
overview from link above:
![2022-10-13_14h16_24](https://user-images.githubusercontent.com/5681880/195624955-095c6fc8-062b-4149-b5bf-c2cfb7a54f2e.png)

hpvd:
more details:
![2022-10-13_14h18_45](https://user-images.githubusercontent.com/5681880/195625079-11db1b5d-b0f3-436b-ae2d-4ee89a5a548c.png)

hpvd:
would/will be very interesting to see, how an updated helm chart with latest pulsar image perform in the same scan.
=> is the update already the big part of the solution?

michaeljmarshall:
@hpvd - thank you for creating this issue. I am going to look into how we can get these issues resolved.

hpvd:
@michaeljmarshall many thanks for starting looking in it that fast
and especially your proposal to make a general approach set the ground to finally solve this (https://github.com/apache/pulsar-helm-chart/pull/299)!

hpvd:
just as background info, the security scanner used by artifacthub, providing results shown above
is trivy, **so all the finding should be pretty valid**.

For details, see:
https://artifacthub.io/docs/topics/security_report/

and trivy
[...]

Issue #65, opened 2020-09-25
Helm chart dependencies do not follow best practices


Imunhatep:
**Is your feature request related to a problem? Please describe.**
Charts do not follow common guideline: chart templates bring dependencies into the main chart instead of using [Helm dependency](https://helm.sh/docs/helm/helm_dependency/)

**Describe the solution you'd like**
Move services like prometheus, alert-manager, zookeeper  to a chart dependencies, That would simplified helm chart support and usage.

**Describe alternatives you've considered**
Do not see other options, cause the issue is that Chart structure is inconsistent and does not follow best practices

**Additional context**
As a chart reference with a lot of dependencies GitLab helm chart could be taken..

chrisjohnson00:
+1
This would also fix the improper use of Prometheus as a deployment.  Prometheus needs to be a stateful set due to it's pvc.

PR #297, opened 2022-10-13
Upgrade to grafana 9.2.0

michaeljmarshall:
Master Issue #294 

### Motivation

The current grafana version is very out of date. We need to upgrade to the latest version. This will trigger a major version bump in the helm chart.

Note that the previous distribution of the Grafana docker image had StreamNative's grafana dashboards. This was convenient for shipping dashboards, but is not the recommended way to deliver dashboards. In the long term, I think we should switch to using the kube prometheus stack and should mount grafana dashboards as config maps. Note that this kind of change is part of modernizing the helm chart and would help address this long standing issue #65.

We'll need to solve this problem before releasing the next version of the helm chart.

Here is a reference to this chart: https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack.

### Modifications

* Upgrade grafana from `streamnative/apache-pulsar-grafana-dashboard-k8s:0.0.16` to `grafana/grafana-oss:9.2.0`.

### Verifying this change

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -189,8 +189,8 @@ images:
     tag: v2.17.2
     pullPolicy: IfNotPresent
   grafana:
-    repository: streamnative/apache-pulsar-grafana-dashboard-k8s
-    tag: 0.0.16
+    repository: grafana/grafana-oss
+    tag: 9.2.0
     pullPolicy: IfNotPresent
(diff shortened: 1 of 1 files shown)


michaeljmarshall:
Superseded by https://github.com/apache/pulsar-helm-chart/pull/299

PR #296, opened 2022-10-13
Upgrade prometheus docker image to 2.39.1

michaeljmarshall:
Master Issue #294 

### Motivation

The current prometheus version is very out of date. We need to upgrade to the latest version. This will trigger a major version bump in the helm chart.

### Modifications

* Upgrade prometheus from 2.12.2 to 2.39.1

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

PR #299, opened 2022-10-14
Replace monitoring solution with kube-prometheus-stack dependency

michaeljmarshall:
Related issues #294 #65

Supersedes #296 and #297

### Motivation

Our helm chart is out of date. I propose we make a breaking change for the monitoring solution and start using the `kube-prometheus-stack` as a dependency. This should make upgrades easier and will let users leverage all of that chart's features.

This change will result in the removal of the StreamNative Grafana Dashboards. We'll need to figure out the right way to address that. The apache/pulsar project has grafana dashboards, but they have not been maintained. With this added dependency, we'll have the benefit of being able to use k8s `ConfigMap`s to configure grafana dashboards.

### Modifications

* Remove old prometheus and grafana configuration
* Add kube-prometheus-stack chart as a dependency
* Enable several components by default. I am not opinionated on these, but it is based on the other values in the chart.

### Verifying this change

This is a large change that will require manual validation, and may break deployments. I propose this triggers a helm chart 3.0.0 release.


hpvd:
just did the same check for the proposed stack at artifacthub.

There are (like in every project) some vulnerabilities, but
- much fewer than we have today
- no damn old ones
- most of them could be fixed by minor version bumps of their dependencies
- and they are moving fast

2022-10-14_10h55_33

see https://artifacthub.io/packages/helm/prometheus-community/kube-prometheus-stack?modal=security-report



-> so it looks fine to use it.

Maybe one could point their community to the general view of their security scan results by opening an issue...

hpvd:
@michaeljmarshall 
since this will be some work and will probably trigger a major version bump,
what do you think of releasing in the meantime a last updated version of 2.x - with no other changes than containing pulsar 2.10.1?

michaeljmarshall:
> @michaeljmarshall since this will be some work and will probably trigger a major version bump, what do you think of releasing in the meantime a last updated version of 2.x - with no other changes than containing pulsar 2.10.1?
> 
> next version with new monitoring stack, could than directly start with pulsar 2.11 as pulsar helm 3.0

I think we need to understand why #274 stalled first. I am working on fixing the release process too, so there will be a slight delay in getting these fixes delivered. However, once the new release process is in place, I expect that we will be able to ship releases more frequently.

michaeljmarshall:
One relevant item to verify is metric label parity. Labels are often added/modified by the relabel config in the scrape definition. 

Metric from publicly available pulsar-helm chart:

> pulsar_broker_lookup{app="pulsar",cluster="p-pulsar",component="broker",controller_revision_hash="p-pulsar-broker-5fc786b7c8",
exported_cluster="p-pulsar",instance="172.17.0.9:8080",job="broker",kubernetes_namespace="default",kubernetes_pod_name="p-pulsar-broker-0",quantile="0.5",release="p",statefulset_kubernetes_io_pod_name="p-pulsar-broker-0"}

Metrics from the helm chart deployed with this PR at this point in git history.

> pulsar_broker_lookup{app="pulsar", cluster="test-pulsar", component="broker", container="test-pulsar-broker", controller_revision_hash="test-pulsar-broker-5d4489dc7d", endpoint="http", exported_cluster="test-pulsar", instance="172.17.0.9:8080", job="broker", kubernetes_namespace="default", kubernetes_pod_name="test-pulsar-broker-0", namespace="default", pod="test-pulsar-broker-0", quantile="0.5", release="test", statefulset_kubernetes_io_pod_name="test-pulsar-broker-0"}

The scrape definitions are pretty different, but ultimately, the broker's current scrape definition is created by the already established `broker-podmonitor.yaml`, so this might be a non-issue.

lhotari (approved):

PR #300, opened 2022-10-14
Remove GitHub Action Workflows that release the chart

michaeljmarshall:
Relates to: https://github.com/apache/pulsar-helm-chart/issues/290

### Motivation

We should not use GitHub Actions to release the helm chart. As such, we can remove the relevant workflow code from this repo while we build the relevant process to officially release the helm chart.

The main risk with this kind of change is that we won't have a way to "release" the chart. However, it is relevant to point out that we have not had any official releases of the chart given that the PMC has not been voting on the releases. I think we need to prioritize fixing this process as a community.

### Modifications

* Remove all scripts and configuration files that enabled GitHub Actions to release the helm chart.

### Verifying this change

This is a trivial change.


dave2wave (approved):

tisonkun:
This PR merged before #274 already causes the Chart index not updated.

michaeljmarshall:
I intentionally merged this before #274 so that it would not trigger an automated "release" of the helm chart. I plan to build a helm chart release candidate on Wednesday or Thursday (CDT).

PR #303, opened 2022-10-17
Allow bk cluster init pod to restart on failure

michaeljmarshall:
### Motivation

This is essentially the same as https://github.com/apache/pulsar-helm-chart/pull/176. Without this change, an init pod can fail and be in `Error` state even though the second pod succeeded. This will prevent misleading errors.

### Modifications

* Replace `Never` with `OnFailure`

### Verifying this change

This is a trivial change.


michaeljmarshall:
Merging now given that it is such a trivial change.

Issue #283, opened 2022-07-28
podManagementPolicy in broker-statefulset


MonicaMagoniCom:
Why in the template broker-statefulset.yaml the value for `podManagementPolicy` if fixed to `Parallel`? Couldn't it be taken from values.yaml?

michaeljmarshall:
@MonicaMagoniCom - it could be taken from the `values.yaml`, yes. Would you like to contribute an enhancement to make it configurable? Thanks!

PR #336, opened 2022-11-23
Zookeeper replicas: adhere to own recommendations

MisterTimn:
The comments say ZK replicacount should be >=3 but defaults to 1, should adhere to own recommendations by default

### Motivation

The comments state >= 3 but the defaults break this recommendation.

### Modifications

Default `zookeeper.replicaCount` is now 3

### Verifying this change

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -281,7 +281,7 @@ zookeeper:
   # so the metrics are correctly rendered in grafana dashboard
   component: zookeeper
   # the number of zookeeper servers to run. it should be an odd number larger than or equal to 3.
-  replicaCount: 1
+  replicaCount: 3
   updateStrategy:
     type: RollingUpdate
   podManagementPolicy: Parallel


charts/pulsar/values.yaml
```diff
@@ -281,7 +281,7 @@ zookeeper:
   # so the metrics are correctly rendered in grafana dashboard
   component: zookeeper
   # the number of zookeeper servers to run. it should be an odd number larger than or equal to 3.
-  replicaCount: 1
+  replicaCount: 3
```
nicoloboschi, line 284: this is a huge breaking change, I don't think we should do it if there's no strong reason
I don't know which comments you're referring to, can you point them?

nicoloboschi (changes requested):
we can't

charts/pulsar/values.yaml
MisterTimn: Line #283 just above the changed line. It states >=3

Zookeeper with 1 replica seems like a bad default either way...

michaeljmarshall: We already provide an example for a minikube deployment https://github.com/apache/pulsar-helm-chart/blob/master/examples/values-minikube.yaml. That configuration overrides this value.

I agree that 3 makes more sense here, and I agree this is essentially a "breaking" change. I would prefer to merge this and go to 3.1.0 and call it out in the release notes.

michaeljmarshall (approved):
LGTM

nicoloboschi (approved):

lhotari (approved):

Issue #246, opened 2022-03-17
Helm Chart requires ClusterRoles for Pulsar Functions, Promotheus, etc. 


briedel:
ClusterRoles for Pulsar Functions, Promotheus, etc. This is related to issue #173 

`rbac.limit_to_namespace: true` 

only limits required pulsar pieces (broker, bookie, etc.) to the namespace, but not Pulsar function workers, Promotheus. 

Error is:

`Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists. Unable to continue with install: could not get information about the resource ClusterRole "test-prep-pulsar-functions-worker" in namespace "": clusterroles.rbac.authorization.k8s.io "test-prep-pulsar-functions-worker" is forbidden: User "..." cannot get resource "clusterroles" in API group "rbac.authorization.k8s.io" at the cluster scope`

`Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists. Unable to continue with install: could not get information about the resource ClusterRole "test-prp-pulsar-prometheus" in namespace "": clusterroles.rbac.authorization.k8s.io "test-prp-pulsar-prometheus" is forbidden: User "..." cannot get resource "clusterroles" in API group "rbac.authorization.k8s.io" at the cluster scope`

lhotari:
The problem is fixed. Please test with latest release of the helm chart.

PR #341, opened 2022-12-05
Respect rbac.limit_to_namespace for Functions RBAC

vfauth:
Fixes #<xyz>

### Motivation

The chart already has a `rbac.limit_to_namespace` parameter to replace ClusterRole(Binding)s with Role(Binding)s. This setting is not used for a ClusterRole required by the functions component, which is an issue for deployments in constrained with limited cluster. See #246 

### Modifications

Handle ClusterRole(Binding) for the functions component the same way as for other components: if `rbac.limit_to_namespace` is `true`, use namespaced Role(Binding) instead.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/broker-rbac.yaml
+++ b/charts/pulsar/templates/broker-rbac.yaml
@@ -19,9 +19,15 @@
 
 {{- if or .Values.components.functions .Values.extra.functionsAsPods }}
 apiVersion: rbac.authorization.k8s.io/v1
+{{- if .Values.rbac.limit_to_namespace }}
+kind: Role
+metadata:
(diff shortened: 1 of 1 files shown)


MarvinCai (approved):

vfauth:
Any news on merging this?

lhotari:
@vfauth I happened to merge #236 which had similar changes. Would you mind checking if the result is desirable? If not, please create a new PR to revisit the solution. One detail is that it would be useful to make the changes backwards compatible so that the roles don't change for existing deployments.

vfauth:
@lhotari They fix the same issue, so you can close mine, thank you.

PR #344, opened 2022-12-17
fix: Set unique advertise address per broker

khrisrichardson:
Fixes #<xyz>

### Motivation

Each but the first broker were failing to register with the cluster. Per `pulsar-admin brokers list use` the first broker was registered with the socket `0.0.0.0:8080` and subsequent brokers attempting to join the cluster failed, since they were competing for the same advertised address.

### Modifications

This PR sets the advertised address uniquely per pod using the Kubernetes downward API to determine the pod's IP.

### Verifying this change


khrisrichardson:
PEBKAC. It looks like I had overridden the advertised address to `0.0.0.0` in an MQTT configuration overlay.

PR #348, opened 2023-02-26
Fix zookeeper annotations

mouchar:
Fixed the case when no ZK annotations are set and `zookeeper.restartPodsOnConfigMapChange: true` helm can not render template

### Motivation
When you don't set any annotations to Zookeeper (they are empty by default) AND if you set
```yaml
zookeeper:
  restartPodsOnConfigMapChange: true
```
in your values.yaml file, the manifest can not be rendered. This commit fixes this condition and also unifies the
the way how the custom annotations are added, with bookkeeper statefulset template.

```
helm template pulsar apache/pulsar --version 3.0.0 --set zookeeper.restartPodsOnConfigMapChange=true  
Error: YAML parse error on pulsar/templates/zookeeper-statefulset.yaml: error converting YAML to JSON: yaml: line 52: did not find expected key

Use --debug flag to render out invalid YAML
```
### Modifications
wrapping `toYaml` filter (that can return empty object `{}`) into `with` context.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/zookeeper-statefulset.yaml
+++ b/charts/pulsar/templates/zookeeper-statefulset.yaml
@@ -46,7 +46,9 @@ spec:
         {{- if .Values.zookeeper.restartPodsOnConfigMapChange }}
         checksum/config: {{ include (print $.Template.BasePath "/zookeeper-configmap.yaml") . | sha256sum }}
         {{- end }}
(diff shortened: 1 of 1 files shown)


mouchar:
ping - any feedback?
Sorry, but I don't know how to "Make sure that the change passes the CI checks." when the Kind cluster doesn't start at all.

PR #349, opened 2023-03-07
Add nodeAffinity to templates

MonicaMagoniCom:
As right now it is not possible to specify node affinity, I've added it in the templates yml.


lhotari (changes requested):
Thanks for the contribution. Please add defaults to values.yaml .

lhotari:
> Thanks for the contribution. Please add defaults and documentation to values.yaml .

@MonicaMagoniCom Any chance to add this part?

PR #61, opened 2020-09-04
anti-affinity rules refactor and default image tags

EladDolev:
Fixes default anit-affinity type not being used

### Motivation

make life easier, fix broken stuff and add the ability to set `topologyKey`

### Modifications

- add a default image specifications for pulsar components
- a small refactor to anti-affinity rules, add `topologyKey`, and have defaults work
- use `publishNotReadyAddresses` correctly, expose it as a parameter, and remove deprecated alpha annotation

### Verifying this change

- [x] Make sure that the change passes the CI checks.

PR #351, opened 2023-03-16
Configure custom topologyKey for podAntiAffinity

sag-tjegorow:
Adds the ability to configure anti-affinity topology key.

### Motivation

Ability to set custom anti-affinity topology key should make it easier to use pulsar-helm-charts in a high availability multi-AZ cluster (we can use zone/region topology key). Also custom topology will allow the use of keys that are non-standard/not-well-known (see https://kubernetes.io/docs/reference/labels-annotations-taints/), the example being "topology.ebs.csi.aws.com/zone`" from EBS (see https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/729), or just being deprecated (like "failure-domain.beta.kubernetes.io/zone") but still used on older k8s clusters.

I realize there are already 2 PRs waiting for the acceptance that kind of do the same thing:
* https://github.com/apache/pulsar-helm-chart/pull/61 -- at this moment it is a bit stale as it was created in September 2020 (and there are more other changes/refactorings).
* https://github.com/apache/pulsar-helm-chart/pull/127 -- functionally very similar to this one (except soft affinity weight configuration) but does not allow to configure custom topology key. Being able to configure the topology key makes this PR simpler.

### Modifications

Just the ability to configure topology keys using new values: <component>.affinity.anti_affinity_topology_key -- defaults to currently hard-coded "kubernetes.io/hostname".

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/autorecovery-statefulset.yaml
+++ b/charts/pulsar/templates/autorecovery-statefulset.yaml
@@ -81,7 +81,7 @@ spec:
                 operator: In
                 values:
                 - {{ .Values.autorecovery.component }}
-            topologyKey: "kubernetes.io/hostname"
+            topologyKey: {{ .Values.autorecovery.affinity.anti_affinity_topology_key }}
         {{ else }}
           {{ .Values.autorecovery.affinity.type }}:
             - weight: 100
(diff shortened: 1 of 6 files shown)


lhotari (approved):
LGTM

PR #354, opened 2023-03-25
Improve HPA

AtkinsChang:
### Motivation

- Support Kubernetes 1.26 `warning "autoscaling/v2beta2 HorizontalPodAutoscaler is deprecated in v1.23+, unavailable in v1.26+; use autoscaling/v2 HorizontalPodAutoscaler"`
- Field `.spec.replicas` should be owned by `kube-controller-manager` when HPA is enabled

### Modifications

- Use `autoscaling/v2` if Kubernetes version >= 1.23
- When `.Values.<component>.autoscaling.enabled` is set to true, field `.spec.replicas` is not generated

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/broker-hpa.yaml
+++ b/charts/pulsar/templates/broker-hpa.yaml
@@ -18,7 +18,11 @@
 #
 
 {{- if .Values.broker.autoscaling.enabled }}
+{{- if (semverCompare "<1.23-0" .Capabilities.KubeVersion.Version) }}
 apiVersion: autoscaling/v2beta2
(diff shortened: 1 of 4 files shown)


lhotari (approved):
LGTM

amritsrivastava:
When are we planning to release this as this is breaking the current Helm releases? 
There is no tag published for this change.

tripolif98:
Are there any updates on releasing a version with this change?
As far as we can tell without this PR released it won't be possible to upgrade Kubernetes clusters to version 1.26

Issue #335, opened 2022-11-16
Proxy fails to connect to port 80


briedel:
When using the helm chart on GKE, the proxy is unable to connect to port 80, which crashes the proxy pod and begins a restart loop. The logs how this

```
Defaulted container "pulsar-300-proxy" out of: pulsar-300-proxy, wait-zookeeper-ready (init), wait-broker-ready (init)
[conf/proxy.conf] Applying config authenticationEnabled = true
[conf/proxy.conf] Applying config authenticationProviders = org.apache.pulsar.broker.authentication.AuthenticationProviderToken
[conf/proxy.conf] Applying config authorizationEnabled = false
[conf/proxy.conf] Applying config brokerClientAuthenticationParameters = file:///pulsar/tokens/proxy/token
[conf/proxy.conf] Applying config brokerClientAuthenticationPlugin = org.apache.pulsar.client.impl.auth.AuthenticationToken
[conf/proxy.conf] Applying config brokerServiceURL = pulsar://pulsar-300-broker:6650
[conf/proxy.conf] Applying config brokerWebServiceURL = http://pulsar-300-broker:8080
[conf/proxy.conf] Applying config clusterName = pulsar-300
[conf/proxy.conf] Applying config forwardAuthorizationCredentials = true
[conf/proxy.conf] Applying config httpNumThreads = 8
[conf/proxy.conf] Applying config servicePort = 6650
[conf/proxy.conf] Applying config statusFilePath = /pulsar/status
[conf/proxy.conf] Applying config superUserRoles = admin,broker-admin,proxy-admin
[conf/proxy.conf] Applying config tokenPublicKey = file:///pulsar/keys/token/public.key
[conf/proxy.conf] Applying config webServicePort = 80
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by io.netty.util.internal.ReflectionUtil (file:/pulsar/lib/io.netty-netty-common-4.1.77.Final.jar) to constructor java.nio.DirectByteBuffer(long,int)
WARNING: Please consider reporting this to the maintainers of io.netty.util.internal.ReflectionUtil
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
2022-11-16T17:17:00,431+0000 [main] INFO  org.apache.pulsar.broker.authentication.AuthenticationService - [org.apache.pulsar.broker.authentication.AuthenticationProviderToken] has been loaded.
[...]

PR #356, opened 2023-03-28
fix: proxy should not use priviledged port numbers

chrisjohnson00:
This fixes issue #335

### Motivation

Fixing the port 80 binding issue in issue #335 

### Modifications

Changed from port 80 -> 8080 for HTTP and 443 -> 8443 for HTTPS

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/proxy-service.yaml
+++ b/charts/pulsar/templates/proxy-service.yaml
@@ -37,17 +37,21 @@ spec:
     - name: http
       port: {{ .Values.proxy.ports.http }}
       protocol: TCP
+      targetPort: sts-http
     - name: "{{ .Values.tcpPrefix }}pulsar"
       port: {{ .Values.proxy.ports.pulsar }}
       protocol: TCP
+      targetPort: "sts-{{ .Values.tcpPrefix }}pulsar"
     {{- end }}
     {{- if and .Values.tls.enabled .Values.tls.proxy.enabled }}
     - name: https
       port: {{ .Values.proxy.ports.https }}
       protocol: TCP
+      targetPort: sts-https
     - name: "{{ .Values.tlsPrefix }}pulsarssl"
       port: {{ .Values.proxy.ports.pulsarssl }}
       protocol: TCP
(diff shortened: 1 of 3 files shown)


michaeljmarshall (approved):
LGTM. Thanks for your contribution @chrisjohnson00!

The change to run the container as a non-root user broke this part of the helm chart. Another option would be to let the proxy service continue using ports 80 and 443 while mapping those to 8080 and 8443 on the proxy container. I don't think that is necessary though if we correctly document this change in the release notes.

chrisjohnson00:
> LGTM. Thanks for your contribution @chrisjohnson00!
> 
> The change to run the container as a non-root user broke this part of the helm chart. Another option would be to let the proxy service continue using ports 80 and 443 while mapping those to 8080 and 8443 on the proxy container. I don't think that is necessary though if we correctly document this change in the release notes.

Oh, I hadn't noticed that both the service and the deployment shared these values.  Wouldn't this make it a breaking change since the service ports would change?  I'd be happy to update it so it is not a BC!

michaeljmarshall:
> Oh, I hadn't noticed that both the service and the deployment shared these values. Wouldn't this make it a breaking change since the service ports would change? I'd be happy to update it so it is not a BC!

Sorry for my delayed response. Yes, it would make it a breaking change. It'd be great if we can avoid the breaking change, thanks!

chrisjohnson00:
> > Oh, I hadn't noticed that both the service and the deployment shared these values. Wouldn't this make it a breaking change since the service ports would change? I'd be happy to update it so it is not a BC!
> 
> Sorry for my delayed response. Yes, it would make it a breaking change. It'd be great if we can avoid the breaking change, thanks!

I've pushed up a change that makes this backward compatible.  Let me know what you think.
I've introduced using named ports for the `targetPort` in the service, which helps to further protect future change from being breaking.  Since the port names in the service and stateful set were the same, I prefixed the ports in the sts with `sts-` to distinguish them.

lhotari:
closing and re-opening to trigger new CI run

PR #367, opened 2023-07-10
Update README.md

geomagilles:
### Motivation

this repo is not actively maintained and has many open issues - it gives a bad impression to a newcomer regarding the Pulsar community.

### Modifications

I propose to remove the "Official" term from the description of this repository and added a warning section to manage expectations.

### Verifying this change

- not needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,13 @@
     under the License.
 
 -->
-# Official Apache Pulsar Helm Chart
+# Apache Pulsar Helm Chart
 
-This is the officially supported Helm Chart for installing Apache Pulsar on Kubernetes.
+This is a Helm Chart for installing Apache Pulsar on Kubernetes.
 
 Read [Deploying Pulsar on Kubernetes](http://pulsar.apache.org/docs/deploy-kubernetes/) for more details.
+
+> :warning: This helm chart is updated outside of the regular Pulsar release cycle and might lag behind a bit. It only supports basic Kubernetes features. Ultimately, it's meant as a template and starting point for a Kubernetes deployment. In many cases, it would require some customizations.
 
 ## Features
 


eolivelli (changes requested):
Why are you dropping "official" ?

lhotari (changes requested):
-1

Let's fix the problems instead. @geomagilles Please engage with the community on dev@pulsar.apache.org mailing list so that we get broader community awareness of the issues.

merlimat:
@eolivelli @lhotari The current problems with the Helm chart are not something that can be easily fixed, unless significant investment is committed to the effort.

In my view: 
 * The Helm chart is lagging behind Pulsar releases with no ETAs on whether the "current" release will be available
 * Many of the feature combinations are not working out of the box, because of changes in K8S APIS, compatibility with different cloud providers, etc.. 
 * The result is that, currently, it's very difficult to get anything other than a trivial basic clusters working out of the box.

As per the source of these issues, I think it comes that mostly everyone uses different means to deploy Pulsar in Kubernetes (eg: operators or forks of the original Helm chart), therefore this Helm Chart is practically untested by most of the people working on the project.

For a user, though, we are promising that this Helm chart is the best preferred way to deploy Pulsar in Kubernetes, while in fact it does not work out of the box for the majority of the users (as you can see from the number of people getting stuck and asking questions in Slack). 

So, again in my view, there are 2 options: 
 1. Set the expectations on the "real" status of this Helm chart
 2. Invest significant resources to make it work well, in the way users would expect from an "officially supported" component. 

Is there anyone who has the bandwidth to commit for making (2) happen?

README.md
```diff
@@ -18,12 +18,14 @@
...
 -->
-# Official Apache Pulsar Helm Chart
+# Apache Pulsar Helm Chart
 
-This is the officially supported Helm Chart for installing Apache Pulsar on Kubernetes.
+This is a Helm Chart for installing Apache Pulsar on Kubernetes.
 
 Read [Deploying Pulsar on Kubernetes](http://pulsar.apache.org/docs/deploy-kubernetes/) for more details.
 
+> :warning: This helm chart is updated outside of the regular Pulsar release cycle and might lag behind a bit. It only supports basic Kubernetes features. Ultimately, it's meant as a template and starting point for a Kubernetes deployment. In many cases, it would require some customizations.
```
tisonkun, line 27:
```suggestion
> :warning: This helm chart is updated outside of the regular Pulsar release cycle and might lag behind a bit. It only supports basic Kubernetes features now. Currently, it can be used as no more than a template and starting point for a Kubernetes deployment. In many cases, it would require some customizations.
```

tisonkun (approved):
I agree that we should update the project status to its real status.

Comments to change the words to reflect the situation instead of making conclusions. If the project revives and can serve most of production usage, the developers can later change the words again.

lhotari:
@merlimat @tisonkun Could either of you bring this issue to the dev mailing list for discussion so that the community is aware of this situation and we collectively decide on the resolution. Thanks!

lhotari:
I'm closing this PR since Apache projects make project decisions on the mailing list.

tisonkun:
I'm going to send a simila PR according to the lazy consensus on https://lists.apache.org/thread/zrqs973cbvm7szs8kktgbgfohs8gdrzp.

As mentioned in the thread:

> I totally appreciate anyone who is willing to maintain the pulsar-helm-chart, but let's do not block the description updates by such an argument. Instead, update the description, and change it back when the development and maintenance really happen. It also reflects the low traffic during the past few years.

PR #375, opened 2023-08-08
Fix: Bookie don't get configured PULSAR_MEM properly from chart

mnit016:
Fixes keytool.sh overwriting PULSAR_EXTRA_OPTS & BOOKIE_EXTRA_OPTS 

### Motivation
Bookie didn't get configured PULSAR_MEM from chart, because [Keytool.sh Line 100](https://github.com/apache/pulsar-helm-chart/blob/73fe688a439c3ab9b56f2d249f16505292391f4b/charts/pulsar/templates/keytool.yaml#L100) echo a line which overwrite the old env.

### Modifications
add a slash in front of $ in order to pass env definition instead of value.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/keytool.yaml
+++ b/charts/pulsar/templates/keytool.yaml
@@ -95,11 +95,11 @@ data:
     
     if [[ "x${isClient}" == "xtrue" ]]; then
         echo $'\n' >> conf/pulsar_env.sh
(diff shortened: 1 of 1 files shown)


lhotari (approved):

lhotari:
Thanks for the contribution @mnit016

PR #376, opened 2023-08-08
Enables Nodeport Support For the Proxy Service

apetti1920:
Adds optional NodePort Support for the proxy service

# Motivation
When creating local kube clusters there are times where nodeports become more useful and easier to setup for testing than setting up the load balancer.

# Modifications
Optional nodeport options were added for the http, https, pulsar, and pulsarssl services via the proxy

# Verifying this change

### DIFF
--- a/charts/pulsar/templates/proxy-service.yaml
+++ b/charts/pulsar/templates/proxy-service.yaml
@@ -38,20 +38,40 @@ spec:
       port: {{ .Values.proxy.ports.http }}
       protocol: TCP
       targetPort: sts-http
+      {{- if and (eq .Values.service.type "NodePort") (ne .Values.service.nodePorts.http "") }}
+        nodePort: {{ .Values.service.nodePorts.http }}
+      {{- else}}
+        nodePort: null
+      {{- end}}
     - name: "{{ .Values.tcpPrefix }}pulsar"
       port: {{ .Values.proxy.ports.pulsar }}
       protocol: TCP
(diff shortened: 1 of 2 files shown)


bnovick:
+1 on this PR , would be excellent to have this flexibility! @michaeljmarshall I've seen you tagged on similar areas , would you mind taking a look?

charts/pulsar/templates/proxy-service.yaml
```diff
@@ -38,20 +38,40 @@ spec:
       port: {{ .Values.proxy.ports.http }}
       protocol: TCP
       targetPort: sts-http
+      {{- if and (eq .Values.service.type "NodePort") (ne .Values.service.nodePorts.http "") }}
+        nodePort: {{ .Values.service.nodePorts.http }}
```
lhotari, line 42: please check the indentation

charts/pulsar/templates/proxy-service.yaml
```diff
@@ -38,20 +38,40 @@ spec:
...
+      {{- else}}
+        nodePort: null
```
lhotari, line 44: Could we simply omit setting nodePort instead of setting the value to null?

lhotari (changes requested):

lhotari:
@apetti1920 Pleaer rebase and address the review comments

lhotari:
covered by #500

PR #381, opened 2023-08-18
Initialize Pulsar cluster with Proxy config 

vineeth1995:
### Motivation

We need to update Pulsar cluster metadata with ProxyUrl while using ATS proxy. So, support Pulsar-cluster-metadata-init with proxy config

### Modifications

Add proxy-url and proxy-protocol values in Pulsar-cluster-metadata-init

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/pulsar-cluster-initialize.yaml
+++ b/charts/pulsar/templates/pulsar-cluster-initialize.yaml
@@ -105,6 +105,12 @@ spec:
               {{- if not .Values.pulsar_metadata.configurationStore }}
               --configuration-store {{ template "pulsar.zookeeper.connect" . }}{{ .Values.metadataPrefix }} \
               {{- end }}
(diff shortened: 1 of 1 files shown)


charts/pulsar/templates/pulsar-cluster-initialize.yaml
```diff
@@ -105,6 +105,12 @@ spec:
               {{- if not .Values.pulsar_metadata.configurationStore }}
               --configuration-store {{ template "pulsar.zookeeper.connect" . }}{{ .Values.metadataPrefix }} \
               {{- end }}
+              {{- if not .Values.pulsar_metadata.proxyUrl }}
```
lhotari, line 108: Does it make sense to have a "not" rule?

lhotari (changes requested):

lhotari:
Closing since there was no response from the contributor for the review.

PR #385, opened 2023-08-23
fix backend boot failure

jak78:
Fixes #505

### Motivation

Using the default settings of the Pulsar Helm Chart to run Pulsar Manager results in a non-functional manager instance. This PR suggests a fix.

### Modifications

Fix `examples/values-minikube.yml` as suggested [here](https://github.com/apache/pulsar-manager/issues/505#issuecomment-1606947653) by @jesperbagge

### Verifying this change

### DIFF
--- a/examples/values-minikube.yaml
+++ b/examples/values-minikube.yaml
@@ -49,3 +49,9 @@ broker:
 
 proxy:
   replicaCount: 1
+
+pulsar_manager:
+  configData:
+    ENV_SPRING_CONFIGURATION_FILE: "/pulsar-manager/pulsar-manager/application.properties"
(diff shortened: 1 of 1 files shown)


jak78:
That should fix the following issue: https://github.com/apache/pulsar-manager/issues/505

Mortom123:
Setting the variable `ENV_SPRING_CONFIGURATION_FILE` is not needed. Its just[ supervisord.d syntax](http://supervisord.org/configuration.html#environment-variables) for accessing the underlying `SPRING_CONFIGURATION_FILE` env variable. 
I stumbled upon the linked issue + fix, however would like to point out that removing the self reference from `PULSAR_MANAGER_OPTS` suffices to fix the issue.

PR #392, opened 2023-09-10
Add function config values to function-worker-configmap.yaml

conor-nsurely:
Previously config values for function were not being added to the K8s configmap even if they were set in the values.yaml under the functions section.

### Motivation
I have been unable to configure Pulsar functions via the values.yaml( E.g. configuring runtimeCustomizerClassName) .
The values I set do not appear in the configmap that is generated by Helm in K8s, and therefore the changes are not reflected in Pulsar.

### Modifications

I have changed the template for the function-worker-configmap.yaml so that it will add the correct config values to the k8s configmap.

### DIFF
--- a/charts/pulsar/templates/function-worker-configmap.yaml
+++ b/charts/pulsar/templates/function-worker-configmap.yaml
@@ -29,4 +29,6 @@ metadata:
     component: {{ .Values.functions.component }}
 data:
   pulsarDockerImageName: "{{ template "pulsar.imageFullName" (dict "image" .Values.images.functions "root" .) }}"
+{{ toYaml .Values.functions | indent 2 }}
+
 {{- end }}


charts/pulsar/templates/function-worker-configmap.yaml
```diff
@@ -29,4 +29,6 @@ metadata:
     component: {{ .Values.functions.component }}
 data:
   pulsarDockerImageName: "{{ template "pulsar.imageFullName" (dict "image" .Values.images.functions "root" .) }}"
+{{ toYaml .Values.functions | indent 2 }}
```
lhotari, line 32: This doesn't look like the right solution. In this helm chart, the function worker runs together with the broker. The configuration in `.Values.broker.configData` with `PF_` prefix would get applied to the function_worker.yml file.

The keys in this file get applied directly in code with a timer, ever 5 minutes with this code https://github.com/apache/pulsar/blob/613a77100226628d8685d34260685d2df2b405ae/pulsar-functions/runtime/src/main/java/org/apache/pulsar/functions/runtime/kubernetes/KubernetesRuntimeFactory.java#L408-L422 to the KubernetesRuntimeFactory instance. 
The solution is buggy and will only work for String fields. Another problem is that initially it will start with the default values and override after running for 5 minutes. Therefore it seems like a useless solution.

lhotari:
will be clarified with #462

PR #404, opened 2023-11-03
feat: add imagePullSecrets in Job

lenglet-k:
Fixes #imagePullSecrets in job

### Motivation

We use private pulsar image on job, we cannot pull this image from our registry because imagePullSecrets is not in job template

### Modifications

i have add imagePullSecrets in pulsar-cluster-initialize.yaml and bookkeeper-cluster-initialize.yaml

### Verifying this change


lhotari (approved):
Thanks for the contribution @lenglet-k . LGTM

PR #410, opened 2023-11-22
Use container ports for proxy stateful set probes

emre-aydin:
### Motivation

When the Helm chart is installed, proxy pods fail to become ready because the wrong ports are used for the probes.

### Modifications

I have simply changed the ports used for probes from the service ports to the container ports.

### Verifying this change


lhotari (approved):

lhotari:
Thank you for the contribution @emre-aydin

PR #421, opened 2023-12-26
Add loadBalancerIP support for proxy service

hadican:
### Motivation

There might be cases where the proxy load balancer IP should be static to whitelist it.

### Modifications

I've simply added `loadBalancerIP` field to the proxy service. It's optional, can be empty.

### Verifying this change


hadican:
@lhotari

PR #432, opened 2024-01-16
Add testing of upgrading of latest released chart version to CI

lhotari:
### Motivation

It is important that new chart version don't introduce breaking changes that prevent upgrading a cluster from an older version to the newest version.

### Modifications

- Remove "basic" test scenario and replace it with "upgrade".
- Add 2 upgrade scenarios for upgrading the PSP and TLS configurations

- Upgrade installs the latest released chart version & produces 2 messages to a Pulsar topic
  - It upgrades the cluster and then consumes the 2 messages

- Adds "metallb" for k8s loadbalancer support so that `--wait` will work with `helm upgrade`. Previous waiting for a deployment to complete never succeeded since the load balancer didn't get an IP. Installing metallb to kind solves that problem.

### Verifying this change


lhotari:
Merging since this is needed for testing other PRs.

PR #435, opened 2024-01-17
Don't use TLS from function instances to brokers by default

lhotari:
### Motivation

- CI is currently broken after enabling Pulsar Function testing in all configuration with #434
- When TLS is enabled Function instances cannot communicate to the broker since Function instances don't currently have the custom TLS CA cert available.
- When the function worker runs on the broker, the configuration `PF_tlsAllowInsecureConnection: "true"` has no impact and it's not possible to configure `tlsAllowInsecureConnection=true` only for the function instance. [The function worker uses the broker configuration for the tlsAllowInsecureConnection setting](https://github.com/apache/pulsar/blob/fa81bbcacb33c57c24f0d4ce82d5e25d3ac1f167/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java#L1820) and changing that would have a larger impact.

### Modifications

- Add new configuration `tls.function_instance.enabled` that defaults to false. This controls whether TLS is used from function instances to the broker.

### Verifying this change


tisonkun (approved):
Thank you!

Issue #426, opened 2024-01-11
Function Worker / Broker RBAC alignment


Mortom123:
**Is your feature request related to a problem? Please describe.**
Right now `broker-rbac.yaml` basically just performs work regarding function workers. This file should be renamed accordingly (`function-worker-rbac.yaml`). 
`broker-cluster-role-binding.yaml` can then be renamed to `broker-rbac.yaml` aligning with the rest of the templates. 

Both, function-workers and broker, are the only ones having `-rbac.yaml`'s then. maybe changing to `psp` is more suited here aswell.

**Describe the solution you'd like**
see above

**Describe alternatives you've considered**
None.

**Additional context**
Willing to submit a PR :)

lhotari:
@Mortom123 Yes, it makes sense what you are proposing. I'm looking forward to your PR.

PR #438, opened 2024-01-21
General Cleanup + Pulsar Manager Initialization from secrets

Mortom123:
This pull request implements #426. It also moves some files around for overall better grouping of resources. 
Additionally, the deployed pulsar-manager is conditioned with given secret credentials so the web login works out of the box. 

### Motivation

The project had some historical growth, such that there exist a lot of template files now each prefixed with a component. This PR aims to increase the overall readability of this repo by splitting these files into different directories. This also makes it easier to get a quick overview which components have a `PodMonitor`, a `ServiceAccount`, ... associated with them.

Setting the password for the pulsar-manager in `values.yaml` does only set the internal database connection of the manager but not the outward-facing website credentials. This is now also implemented. 

### Modifications
- moved files around
- created an additional Batch Job which is along the lines of the `pulsar-cluster-initialize.yaml` but conditions the pulsar-manager, for this a new template function for the secret is introduced to keep it DRY. 
- exposed the admin port of the pulsar-manager in its service
- renamed the initContainer resources for the nslookup container to better reflect where they are used

### Verifying this change


Mortom123:
@lhotari the license issue is [fixed](https://github.com/Mortom123/pulsar-helm-chart/actions/runs/7602852948).

lhotari:
Thanks for the contribution @Mortom123 ! It looks like there's a lot of good improvements packaged here.

> This pull request implements #426. It also moves some files around for overall better grouping of resources.
> Additionally, the deployed pulsar-manager is conditioned with given secret credentials so the web login works out of the box.

Please create individual PRs for separate changes. You don't have to create matching issues for each PR, but that can be done if that helps.

> * moved files around

This change is something that would be great to do, but it causes a significant drawback: it breaks diffs for Helm chart users that have modified their own Helm chart and keep it up-to-date by comparing their version to the upstream version. This is a common use case for Helm charts and that's why I'd like to push back on moving all files around.

> * created an additional Batch Job which is along the lines of the `pulsar-cluster-initialize.yaml` but conditions the pulsar-manager, for this a new template function for the secret is introduced to keep it DRY.

This is the main change for Pulsar Manager related changes?

> * exposed the admin port of the pulsar-manager in its service

I guess the admin port change could be a separate PR too

> * renamed the initContainer resources for the nslookup container to better reflect where they are used

This is a separate change so it should be handled in a separate PR.

lhotari:
> Setting the password for the pulsar-manager in `values.yaml` does only set the internal database connection of the manager but not the outward-facing website credentials. This is now also implemented.

It might be better to use separate credentials for the database and the website login?

Mortom123:
@lhotari Admittedly, I thought about that but decided not to do so, because I would say this simply adds "configuration overhead". Afaik, if you have either credentials (website/db) you can control the other, so I opted to keep them aligned. But if you think its better otherwise I can change that.

lhotari:
> @lhotari Admittedly, I thought about that but decided not to do so, because I would say this simply adds "configuration overhead". Afaik, if you have either credentials (website/db) you can control the other, so I opted to keep them aligned. But if you think its better otherwise I can change that.

I think it's not good to do shortcuts with security. Another part to verify is that Pulsar Manager doesn't use a default password as it does today. A random password should be the default password. Found this on Twitter: https://twitter.com/fattselimi/status/1648415859453435904 . This is not the first time I've seen such tweets.

One approach would be to generate a secret at installation time if the secret doesn't exist. Helm can check existence of Kubernetes objects [with the `lookup` function](https://helm.sh/docs/chart_template_guide/function_list/#kubernetes-and-chart-functions) and add add `"helm.sh/resource-policy": "keep"` annotation.
(somewhat similar ideas here, but slightly different: https://itnext.io/manage-auto-generated-secrets-in-your-helm-charts-5aee48ba6918)

> I agree. Moving files around also breaks a lot of forks / pending PR's. Maybe for the next major release we can opt to do so?

Yes, it would be possible in the next major release.

> 
> I can go on as follows:
> 
> 1. PR: rename lookup ressources
> 2. PR: expose Admin Port of Manager
> 3. PR: `pulsar-cluster-initialize.yaml`  (maybe an additional one before that introducing the `_manager.tpl` file?)
> 4. PR: move functions worker / broker service account around

Yes, makes sense.

PR #440, opened 2024-01-23
expose admin port of pulsar manager in service

Mortom123:
See https://github.com/apache/pulsar-helm-chart/pull/438 / https://github.com/apache/pulsar-helm-chart/pull/438#issuecomment-1903889439  2.PR

### Motivation

exposing Admin Port of Manager is needed in the future

### Modifications

expose Admin Port of Manager

### Verifying this change

- [x] Make sure that the change passes the CI checks.  See https://github.com/Mortom123/pulsar-helm-chart/actions/runs/7629866629

### DIFF
--- a/charts/pulsar/templates/pulsar-manager-deployment.yaml
+++ b/charts/pulsar/templates/pulsar-manager-deployment.yaml
@@ -64,6 +64,7 @@ spec:
         {{- end }}
           ports:
           - containerPort: {{ .Values.pulsar_manager.service.targetPort }}
+          - containerPort: {{ .Values.pulsar_manager.service.adminTargetPort }}
           volumeMounts:
           - name: pulsar-manager-data
             mountPath: /data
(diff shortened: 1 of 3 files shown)


lhotari (approved):

lhotari:
I already merged this, but I do have a concern. Since the admin port was added to the existing service that is a load balancer, I guess this expands the attack surface since admin operations are now accessible on the load balancer. @Mortom123 any thoughts on that?

lhotari:
> @lhotari Mhh. You might be right. However I think the admij port gets deactivated / does not listen anymore once the account for the manager is created. But we could also make a second Node Port service I guess?

Yes I think it's better to reduce to possible exposure. One fact is that the Pulsar Helm Chart isn't following a "secure-by-default" principle currently. Anyhow we should attempt to improve the situation. I created #444 to cover other services, but I think that this admin port case could be handled outside of that.

PR #441, opened 2024-01-23
grouped ns lookup containers

Mortom123:
See https://github.com/apache/pulsar-helm-chart/pull/438 / https://github.com/apache/pulsar-helm-chart/pull/438#issuecomment-1903889439  1.PR

### Motivation

rename init container resources to better reflect its actual use (nslookup)

### Modifications

rename init container resources

### Verifying this change

- [x] Make sure that the change passes the CI checks.  See https://github.com/Mortom123/pulsar-helm-chart/actions/runs/7629639467

### DIFF
--- a/charts/pulsar/templates/autorecovery-statefulset.yaml
+++ b/charts/pulsar/templates/autorecovery-statefulset.yaml
@@ -111,7 +111,7 @@ spec:
       - name: pulsar-bookkeeper-verify-clusterid
         image: "{{ template "pulsar.imageFullName" (dict "image" .Values.images.autorecovery "root" .) }}"
         imagePullPolicy: {{ .Values.images.autorecovery.pullPolicy }}
-        resources: {{ toYaml .Values.initContainer_resources.verify_cluster_id | nindent 10 }}
+        resources: {{ toYaml .Values.initContainer.resources | nindent 10 }}
         command: ["sh", "-c"]
         args:
         - >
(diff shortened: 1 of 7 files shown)


lhotari:
> rename init container resources to better reflect its actual use (nslookup)

Perhaps we could come up with a better name? `nslookup` is the command that happens to be used. It would be better to have a name that matches what is being done. The old names reflected this, but we could abstract it to some level, let's say that it's about waiting for a DNS name to become available.
`waitDnsNameAvailable` (following [Helm naming conventions](https://helm.sh/docs/chart_best_practices/values/#naming-conventions))?

Mortom123:
I agree. I also thought about unifying all init container resources. But some start JVM processes, afaik. So this might not be a good approach. What do you think?

lhotari:
> I agree. I also thought about unifying all init container resources. But some start JVM processes, afaik. So this might not be a good approach. What do you think?

I guess that in most cases there isn't a need to tune the init container resources and it's not a concern for most users. A configuration structure with a common setting could be simpler to adjust? 
Keeping things simple should be a guideline also here. Since init containers are short living, there's not necessarily a need to have minimal resources whether it's an init container that uses a JVM or not. Having sufficient resources for all use cases would be needed if there's a single shared config. If we go down the path of a single setting, let's do that. 
I don't think that we should have non-JVM or JVM init container settings. Let's keep the current solution instead of doing that.

Mortom123:
@lhotari I merged all resources

lhotari (approved):
LGTM

PR #448, opened 2024-01-28
Pulsar Manager Initialization and Securing of Admin Secrets

Mortom123:
See https://github.com/apache/pulsar-helm-chart/pull/438 / https://github.com/apache/pulsar-helm-chart/pull/438#issuecomment-1903889439  3.PR

Also the Pulsar-Manager Admin interface is exposed on a different, non outward facing, service because of [given concerns](https://github.com/apache/pulsar-helm-chart/pull/440#issuecomment-1909602387). 

### Motivation

Right now you need to initialize the pulsar-manager manually, this can be done as an init job. 

### Modifications

1. Change Admin Port exposure of Pulsar Manager to second internal service
2. Provide means to securely generate a password / use an existing one
3. Provide a job that runs on init and conditions the Pulsar Manager

### Verifying this change

- [x] Make sure that the change passes the CI checks. [As of right now still running](https://github.com/Mortom123/pulsar-helm-chart/actions/runs/7687391348)

Some things I also noted: 
1. I tried all combinations of BROKER_URL and BOOKIE_URL such that BOOKIE_URL references the Bookie service and one of its ports (as given in the values). However, none of them worked. I suspect there's a naming error on side of the Pulsar Manager
2. There are some errors when starting the manager. Maybe part of this deployment is incorrect.
```
/pulsar-manager/startup.sh: 21: initdb: not found 
/pulsar-manager/startup.sh: 22: pg_ctl: not found 
```

### DIFF
--- a/.ci/chart_test.sh
+++ b/.ci/chart_test.sh
@@ -33,6 +33,8 @@ source ${PULSAR_HOME}/.ci/helm.sh
 # create cluster
 ci::create_cluster
 
+ci::helm_repo_add
+
 extra_opts=""
 if [[ "x${SYMMETRIC}" == "xtrue" ]]; then
     extra_opts="-s"
--- a/.ci/helm.sh
+++ b/.ci/helm.sh
@@ -81,6 +81,13 @@ function ci::install_cert_manager() {
     echo "Successfully installed the cert manager."
 }
 
+function ci::helm_repo_add() {
+    echo "Adding the helm repo ..."
+    ${HELM} repo add prometheus-community https://prometheus-community.github.io/helm-charts
+    ${HELM} repo update
+    echo "Successfully added the helm repo."
+}
+
 function ci::print_pod_logs() {
     echo "Logs for all pulsar containers:"
     for k8sobject in $(${KUBECTL} get pods,jobs -n ${NAMESPACE} -l app=pulsar -o=name); do
--- a/.github/workflows/pulsar-helm-chart-ci.yaml
+++ b/.github/workflows/pulsar-helm-chart-ci.yaml
@@ -127,23 +127,29 @@ jobs:
             --validate-maintainers=false \
             --target-branch ${{ github.event.repository.default_branch }}
 
-      - name: Run kubeconform check
+      - name: Run kubeconform check for helm template with every major k8s version 1.21.0-1.29.0
         if: ${{ steps.check_changes.outputs.docs_only != 'true' }}
         run: |
           PULSAR_CHART_HOME=$(pwd)
           source ${PULSAR_CHART_HOME}/hack/common.sh
+          source ${PULSAR_CHART_HOME}/.ci/helm.sh
+          hack::ensure_kubectl
+          hack::ensure_helm
           hack::ensure_kubeconform
+          ci::helm_repo_add
+          helm dependency build charts/pulsar
           validate_helm_template_with_k8s_version() {
             local kube_version=$1
             echo "Validating helm template with kubeconform for k8s version $kube_version"
             helm template charts/pulsar --set kube-prometheus-stack.enabled=false --set components.pulsar_manager=true --kube-version $kube_version | \
               kubeconform -schema-location default -schema-location 'https://raw.githubusercontent.com/datreeio/CRDs-catalog/main/{{.Group}}/{{.ResourceKind}}_{{.ResourceAPIVersion}}.json' -strict -kubernetes-version $kube_version -summary
           }
-          helm dependency build charts/pulsar
           set -o pipefail
-          validate_helm_template_with_k8s_version 1.21.0
-          validate_helm_template_with_k8s_version 1.27.0
-
+          for k8s_version_part in {21..29}; do
+            k8s_version="1.${k8s_version_part}.0"
+            echo "Validating helm template with kubeconform for k8s version $k8s_version"
+            validate_helm_template_with_k8s_version $k8s_version
+          done
       - name: Wait for ssh connection when build fails
         # ssh access is enabled for builds in own forks
         uses: ./.github/actions/ssh-access


Mortom123:
Please have a look @lhotari, this also addresses some of the issues you mentioned regarding the admin port exposure. Thanks

charts/pulsar/templates/pulsar_manager_cluster_initialize.yaml
```diff
@@ -0,0 +1,137 @@
...
+                -sS -D headers.txt \
+                -d '{"username": "'${USERNAME}'", "password": "'${PASSWORD}'"}'
+
+              LOGIN_TOKEN=$(grep "token:" headers.txt | sed 's/^.*: //')
+              LOGIN_JSESSSIONID=$(grep -o "JSESSIONID=[a-zA-Z0-9_]*" headers.txt | sed 's/^.*=//')
+
+              {{/* create environment */}}
+              {{- if or (not .Values.tls.enabled) (not .Values.tls.broker.enabled) }}
+              BROKER_URL="http://{{ template "pulsar.fullname" . }}-{{ .Values.broker.component }}:{{ .Values.broker.ports.http }}"
+              BOOKIE_URL="http://{{ template "pulsar.fullname" . }}-{{ .Values.broker.component }}:{{ .Values.broker.ports.pulsar }}"
```
Mortom123, line 112: I also tried to use .Values.bookie.component with any of the ports under .Values.bookie.ports, but none worked. However, .Values.bookie.component worked with .Values.broker.ports.pulsar ???

charts/pulsar/values.yaml
```diff
@@ -1326,11 +1327,13 @@ pulsar_manager:
...
     path: "/"
 
-  ## If set use existing secret with specified name to set pulsar admin credentials.
+  ## If set use an existing secret with specified name to set pulsar admin credentials
+  ## Leave empty to generate either a new secret with random passwords or reuse one from a previous installation
+  ## You can find out the actual password by running the following command:
+  ## kubectl get secret -l app=pulsar-manager -o=jsonpath="{.items[0].data}" | base64 --decode
+  ## And decode any key by using:
+  ## kubectl get secret -l app=pulsar-manager -o=jsonpath="{.items[0].data.KEY}" | base64 --decode
   existingSecretName:
```
lhotari, line 1336: is the value of `existingSecretName` referenced anywhere in the template anymore?

charts/pulsar/templates/pulsar_manager_cluster_initialize.yaml
```diff
@@ -0,0 +1,137 @@
...
+              ADMIN_URL={{ template "pulsar.fullname" . }}-{{ .Values.pulsar_manager.component }}-admin:{{ .Values.pulsar_manager.adminService.port }}
+              until $(curl -sS --fail -X GET http://${ADMIN_URL} > /dev/null 2>&1); do
+                  sleep 3;
+              done;
+        # This init container will wait for at least one broker to be ready before
+        # initializing the pulsar-manager
+        - name: wait-broker-ready
+          image: "{{ template "pulsar.imageFullName" (dict "image" .Values.images.proxy "root" .) }}"
+          imagePullPolicy: {{ .Values.images.proxy.pullPolicy }}
+          resources: {{ toYaml .Values.initContainer.resources | nindent 10 }}
```
lhotari, line 64: error message:
```suggestion
          resources: {{ toYaml .Values.initContainer.resources | nindent 12 }}
```
```
W0129 15:44:25.424470   81383 warnings.go:70] unknown field "spec.template.spec.initContainers[1].requests"
```

charts/pulsar/templates/pulsar_manager_cluster_initialize.yaml
lhotari, line 112:
```suggestion
              BOOKIE_URL="http://{{ template "pulsar.fullname" . }}-{{ .Values.bookkeeper.component }}:{{ .Values.bookkeeper.ports.http }}"
```

lhotari: I don't see how it could work with broker's component and port. More comments below.

lhotari (changes requested):
Please address the review comments.

lhotari:
Thanks for the contribution @Mortom123 ! This will be a great improvement to Apache Pulsar Helm chart!

charts/pulsar/values.yaml
Mortom123: Yes. Setting "existingSecretName" basically disables the creation of a secret and lets the application installer inject their own credentials by creating a secret beforehand. Because the section below which was used to enable setting username / password is removed now, this is needed. We could also remove it and use a section in the values to define username/password.

Mortom123: After thinking about it - if I were the end user installing the pulsar helm chart, I would probably prefer tuning some values instead of deploying a secret. So putting it into there might also be a good choice.
However, if the enduser uses something that autogenerates secrets from an external application, providing a Secret name might also be needed.

So we have these decision to make:
1. provide the user with UI_USERNAME / UI_PASSWORD options in the values.yaml file to set their given credentials, yes / no?
2. provide the user with the possibility to overwrite the default secret name with `existingSecretName`, meaning he needs to inject the secret. yes / no?

My choice would probably be:
1. yes
2. no

Mortom123:
@lhotari I updated with your feedback and decided to also reduce the complexity of the secret by only use a single username/password for the UI / DB. I don't see much advantage of having separate ones there.

charts/pulsar/templates/pulsar_manager_cluster_initialize.yaml
```diff
@@ -0,0 +1,137 @@
...
+metadata:
+  name: "{{ template "pulsar.fullname" . }}-{{ .Values.pulsar_manager.component }}-init"
+  namespace: {{ template "pulsar.namespace" . }}
+  labels:
+    {{- include "pulsar.standardLabels" . | nindent 4 }}
+    component: {{ .Values.pulsar_manager.component }}-init
+spec:
+  # This feature was previously behind a feature gate for several Kubernetes versions and will default to true in 1.23 and beyond
+  # https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
+  {{- if .Values.job.ttl.enabled }}
```
lhotari, line 33: I wonder if it would make sense to explicit check for Kubernetes version and enable this by default on k8s 1.23+?
For example:
```
{{- if .Values.job.ttl.enabled or (semverCompare ">=1.23-0" .Capabilities.KubeVersion.Version) }}
  ttlSecondsAfterFinished: {{ .Values.job.ttl.secondsAfterFinished | default 600 }}
{
```

lhotari:
> I updated with your feedback and decided to also reduce the complexity of the secret by only use a single username/password for the UI / DB. I don't see much advantage of having separate ones there.

I think having a single username/password for both UI & DB is simply a very bad practice. It's better to have separate passwords. It's not that much additional overhead to add it now and follow good practices.

> I allowed for manual overwrite of secret values, similar to kube-prometheus-stack, but the default (empty password) is a randomly generated password.

That part looks really good!

> Maybe you have an idea how can test if the login/checking of the environment of the pulsar manager works in the CI pipeline?

Perhaps a simple HTTP request with curl & grep would do the job? That could be added in a similar way as the Pulsar Functions tests:
https://github.com/apache/pulsar-helm-chart/blob/24b80c19865e2d9e0285c54dca05835bb783c8ee/.ci/chart_test.sh#L77-L80
You would add this to `.ci/chart_test.sh` and `.ci/helm.sh` files. In addition there would need to be a test scenario in the GitHub Actions workflow matrix configured in https://github.com/apache/pulsar-helm-chart/blob/master/.github/workflows/pulsar-helm-chart-ci.yaml#L170 with respective values file in .ci/clusters directory. 
When you run the CI jobs in your own fork (by enabling GitHub Actions) and opening a PR to your own fork, you can debug GitHub Actions by logging in by ssh. ssh access is only for your ssh public keys registered in GH, it's https://github.com/Mortom123.keys for you. There's a step in the build that will print the command for connecting with ssh. [example from a build in my fork](https://github.com/lhotari/pulsar-helm-chart/actions/runs/7670897358/job/20908119338#step:5:156):
image
Once you have ssh access, you can debug things to see what problems there are in the CI environment.

> I removed existingSecretName, because when injecting Secrets through secondary applications, I would argue that you have to provide a name for it either way. so might as well set it to the required name.

+1, pulsar-manager integration has been poor so I'm not concerned about the possible breaking change.

> During this I also wondered why we have such flexibility for the names for of the components. Most of the things are setup using pulsar.fullname-component.name. Why don't we fix component.name and only allow for a changing fullname through e.g. other release names?

I guess this is some of the legacy baggage of the chart. As long as there's a backwards compatible way to improve things gradually, we can do that.

charts/pulsar/values.yaml
```diff
@@ -1326,11 +1327,14 @@ pulsar_manager:
...
+  ## On first install, the helm chart tries to reuse an existing secret with matching name by default
+  ## if this should fail it uses the given username and password to create a new secret
+  ## if either are missing the default value of "pulsar" is used for the username or a random password is generated
+  ## And decode any key by using:
+  ## kubectl get secret -l component=pulsar-manager -o=jsonpath="{.items[0].data.PASSWORD}" | base64 --decode
   admin:
-    user: pulsar
-    password: pulsar
+    username: "pulsar"
+    password: "" # leave empty for random password
```
lhotari, line 1337: Linting failed on this line: `1337:18 [comments] too few spaces before comment`

charts/pulsar/templates/pulsar_manager_cluster_initialize.yaml
Mortom123: Sure. I justed copied from the `pulsar-cluster-initalize.yaml`

Mortom123:
@lhotari I added both credentials as well as CI tests. But the tests have some problems:
1. [Here](https://github.com/Mortom123/pulsar-helm-chart/actions/runs/7729592231/job/21073185950) the pulsar manager is not ready to accept connections before being hit by the `curl`. [I tried to fix this](https://github.com/apache/pulsar-helm-chart/pull/448/commits/8d51e36e80362916d3ce91fa8ee38b1891d9503e) by always waiting for jobs to complete in the pipeline. I don't know why we don't wait when installing anyway? Maybe you have some explanation?
2. This also [failed](https://github.com/Mortom123/pulsar-helm-chart/actions/runs/7730062235/job/21074647869), because it takes some time for the job to be triggered (both manager + broker need to be fully ready).

There are now a couple of ways to fix this: 
1. Increase wait timeout in $install_args
2. Revert the commit touching $install args and put an `until` loop before hitting the manager in a similar fashion as [here](https://github.com/apache/pulsar-helm-chart/pull/448/files#diff-74de41a8d85b68c1af8abd4b3d74957919cb9db860425acf0de6a87460719162R53).

I would prefer to wait for all jobs to complete during install. But what do you or @frankjkelly think?

lhotari:
Closing and re-opening to trigger a new build

lhotari:
Closing and re-opening once again to get a fix for the linting job in CI.

Mortom123:
@lhotari i think the build wont work yet. because the job that initializes the manager is not run in timed / waited for, see https://github.com/apache/pulsar-helm-chart/pull/448#issuecomment-1919611961

lhotari:
> 1. I don't know why we don't wait when installing anyway? Maybe you have some explanation?

I guess the reason has been that `--wait` didn't work before adding MetalLB in a8c7745b3ed2cd76a62c81e8c84e3138fd8a3199 . Services of `type: LoadBalancer` require a load balancer implementation in the k8s cluster. MetalLB can be used for "kind" which is used in CI.

Another reason for not using `--wait` might have been that it's possible to have more logging about k8s events and pod state when the helm install command returns and the waiting loop can do whatever it wants. That's just a guess that I'm making.

lhotari:
@Mortom123 if you'd like to debug the issue, it's best to create a PR to your own fork since that will activate ssh access to the build VM. 

https://github.com/apache/pulsar-helm-chart/blob/1f20887f096cf098357c398a714d8db20dc4d551/.github/workflows/pulsar-helm-chart-ci.yaml#L255

It requires a pull request currently and won't work for manually triggered jobs in your fork.

lhotari:
I wonder if upgrading the image to v0.4.0 would help? 
https://github.com/apache/pulsar-helm-chart/blob/1f20887f096cf098357c398a714d8db20dc4d551/charts/pulsar/values.yaml#L187

Mortom123:
Alright. I will try to do so. :) @lhotari

lhotari:
@Mortom123 have you figured out how to use the ssh access to the build? that makes the feedback loop much faster when you have a shell inside a build VM and you can try out different things and debug issues directly. For example, there's `k9s` available to view the kind k8s cluster.

Mortom123:
I will try debugging on my Repo for now and submit a clean PR once its done.

PR #453, opened 2024-01-30
rbac toggles broker roles / rolebindings

Mortom123:
This implements the optional creation of roles/rolebinding for the broker. All other PSP-related resources were previously already toggleable, now broker is also toggleable. 
However, I'm unsure if this has implications on the functionality of Pulsar as a whole.

### Motivation
Some cluster settings prohibit the creation of roles/rolebindings (e.g. Rancher), therefore settings rbac.enabled = False should also prevent the creation of any of these resources. 

### DIFF
--- a/charts/pulsar/templates/broker-cluster-role-binding.yaml
+++ b/charts/pulsar/templates/broker-cluster-role-binding.yaml
@@ -17,7 +17,7 @@
 # under the License.
 #
 
-{{- if .Values.components.broker }}
+{{- if and .Values.components.broker .Values.rbac.enabled }}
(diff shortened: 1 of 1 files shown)


lhotari:
> Some cluster settings prohibit the creation of roles/rolebindings (e.g. Rancher)

Are namespace level roles / rolebinding also prohibited? What is the cluster setting that prevents this? I'm aware that OpenShift and Tanzu are usually handled with namespace level permissions so ClusterRole/ClusterRoleBindings don't work.

Mortom123:
Yes. Our cluster settings prohibit everything role related (cluster or not). They can be created, but only from admin users.

PR #454, opened 2024-01-31
Improve Bookkeeper default configuration

lhotari:
### Motivation

The default Bookkeeper config could be improved so that we let BK choose the defaults for memory settings and that
the data compaction settings are more optimal for a high throughput environment.

### Modifications

- remove minimal memory settings
- add more optimal data compaction settings

### Verifying this change


merlimat (approved):

frankjkelly (commented):
Nice one!

PR #480, opened 2024-03-22
making .ReleaseIsInstall optional for init jobs

### LINKED ISSUE #479: Make it possible to optionally override .ReleaseIsInstall logic in init jobs
In solutions like ArgoCD which don't actually track helm state, .ReleaseIsInstall always evaluates to true.  We should be able to optionally not use .ReleaseIsInstall to determine whether init jobs should be created/removed
An optional parameter to disable using .ReleaseIsInstall to determine if init jobs should be created.

**Additional context**
I will be opening a PR with the necessary changes.

doug-ba:
Fixes #479 

### Motivation

In solutions like ArgoCD which don't actually track helm state, .ReleaseIsInstall always evaluates to true. We should be able to optionally not use .ReleaseIsInstall to determine whether init jobs should be created/removed

### Modifications

makes a new value `useReleaseStatus` which defaults to true.
When `useReleaseStatus` is set to false, the chart relies only on the value of the `initialize` value to determine whether the init jobs should be created.

This PR also changes an `or`  in the pulsar-manager-cluster-initialize template to an `and`.  With the `or` in place, the value passed at `.Values.job.ttl.enabled` was ignored for all clusters newer than Kubernetes version 1.23.0.  This causes the job to be constantly recreated in argocd.
### Verifying this change

### DIFF
--- a/charts/pulsar/templates/bookkeeper-cluster-initialize.yaml
+++ b/charts/pulsar/templates/bookkeeper-cluster-initialize.yaml
@@ -16,7 +16,7 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-{{- if or (and .Values.useReleaseStatus (or .Release.IsInstall .Values.initialize)) .Values.initialize }}
+{{- if or (and .Values.useReleaseStatus .Release.IsInstall) .Values.initialize }}
 {{- if .Values.components.bookkeeper }}
 apiVersion: batch/v1
 kind: Job
--- a/charts/pulsar/templates/bookkeeper-cluster-initialize.yaml
+++ b/charts/pulsar/templates/bookkeeper-cluster-initialize.yaml
@@ -16,7 +16,7 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-{{- if or .Release.IsInstall .Values.initialize }}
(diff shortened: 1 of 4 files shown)


charts/pulsar/templates/bookkeeper-cluster-initialize.yaml
```diff
@@ -16,7 +16,7 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-{{- if or .Release.IsInstall .Values.initialize }}
+{{- if or (and .Values.useReleaseStatus (or .Release.IsInstall .Values.initialize)) .Values.initialize }}
```
lhotari, line 19: It seem that the condition could be simplified 
```suggestion
{{- if or (and .Values.useReleaseStatus .Release.IsInstall) .Values.initialize }}
```

lhotari (commented):
Thanks for the contribution! Really helpful. 

I made a suggestion to simplify the condition. Please check if it makes sense.

charts/pulsar/templates/bookkeeper-cluster-initialize.yaml
doug-ba: Thanks for the feedback, update pushed.

PR #483, opened 2024-03-26
Disable functions by default in values.yaml

lhotari:
### Motivation

Pulsar Functions impose a significant security risk and the default configuration for Pulsar Functions assumes trusted users. 
Pulsar Functions shouldn't be enabled by default.

### Modifications

- Disable Pulsar Functions by default

### Verifying this change

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -120,7 +120,9 @@ components:
   # broker
   broker: true
   # functions
-  functions: true
+  # WARNING! Before enabling functions, make sure that all of your users are trusted since functions run user code
+  # and the current security sandbox is not sufficient to protect against malicious code.
+  functions: false
   # proxy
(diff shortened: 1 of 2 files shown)


charts/pulsar/values.yaml
```diff
@@ -120,7 +120,9 @@ components:
   # broker
   broker: true
   # functions
-  functions: true
+  # WARNING! Before enabling functions, make sure that all of your users are trusted since functions run user code
+  # and the current security sandbox is not sufficient to protect against malicious code.
+  functions: false
```
frankjkelly, line 125: Is there any documentation for function security that they could be pointed to?

lhotari: No such documentation exists currently.

eolivelli (approved):
+1

PR #485, opened 2024-04-03
Bugfix/pulsar manager persistence

### LINKED ISSUE #470: Pulsar Manager v0.4.0 in kubernetes still can not persist any data
**Describe the bug**
I upgraded our Pulsar to Chart v3.3.0, and the custom values for Pulsar Manager are as follows:
```yml
pulsar_manager:
  nodeSelector:
    pulsar: "true"
  service:
    type: ClusterIP
  volumes:
    data:
      storageClassName: ebs-gp3
      local_storage: false
      size: 1Gi
```
After creating an environment and some tenants in the web console, I restarted the Pulsar Manager pod to verify whether the data would be persisted. The result was that no data was persisted. I had to recreate the admin user and environment via the REST API.

In the Pulsar Manager POD, the configuration file (/pulsar-manager/pulsar-manager/application.properties) defaults to using HerdDB (which is not persisted), and the PostgreSQL related flags were commented out:
```ini
spring.datasource.schema=classpath:/META-INF/sql/herddb-schema.sql
spring.datasource.username=sa
spring.datasource.password=hdb
spring.datasource.initialization-mode=always

# postgresql configuration
#spring.datasource.driver-class-name=org.postgresql.Driver
#spring.datasource.url=jdbc:postgresql://127.0.0.1:5432/pulsar_manager
#spring.datasource.username=postgres
#spring.datasource.password=postgres
```
In the PV, I only found files of postgresql database, but no herddb relates files.

**To Reproduce**
1. install the latest version of Chart will  reproduce it:
`helm install pulsar pulsar/pulsar`

Mortom123:
Fixes #470 

### Motivation
Reconditioning the pulsar manager after reinstall should not be necessary and data should be persisted.

### Modifications
Multiple modifications. It did not work beforehand because `SPRING_CONFIGURATION_FILE` was set in the environment resulting in a default configuration.

During my investigations, I found out that the **pulsar manager does not setup the database correctly and instead defaults to user/password = pulsar/pulsar**. 
Once the changes on the side of the manager are implemented the same credentials as specified should be used. 
See: https://github.com/apache/pulsar-manager/pull/556

### Verifying this change

### DIFF
--- a/.ci/helm.sh
+++ b/.ci/helm.sh
@@ -357,8 +357,7 @@ function ci::test_pulsar_manager() {
   until ${KUBECTL} get jobs -n ${NAMESPACE} ${CLUSTER}-pulsar-manager-init -o json | jq -r '.status.conditions[] | select (.type | test("Complete")).status' | grep True; do sleep 3; done
   ${KUBECTL} describe job -n ${NAMESPACE} ${CLUSTER}-pulsar-manager-init
   ${KUBECTL} logs -n ${NAMESPACE} job.batch/${CLUSTER}-pulsar-manager-init
-  # this line errors in some tests? -  i do not know why, but is really useful for debugging, try: cat ./pulsar-manager.log otherwise
-  # ${KUBECTL} exec -n ${NAMESPACE} ${CLUSTER}-pulsar-manager-0 -- cat /pulsar-manager/pulsar-manager/pulsar-manager.log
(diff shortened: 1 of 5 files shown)


Mortom123:
I think this solves the previous issues @lhotari @wuyudian1

lhotari:
Awesome work @Mortom123 ! Thank you

lhotari (approved):

PR #509, opened 2024-06-27
OpenID: introducing support for OpenID configuration

brunodomenici:
BREAKING CHANGE: provider configuration changed from auth.authentication.provider to auth.authentication.jwt.enabled

### Motivation

Allow the configuration of the OpenID authentication provider along side with JWT.

### Modifications

Provider configuration `auth.authentication.provider` (only 'jwt' value was accepted) was decommissioned in favor of `auth.authentication.jwt.enabled`:

From:

```yaml
auth:
  authentication:
    enabled: false
    provider: "jwt"
```

To:

```yaml
# Enable or disable broker authentication and authorization.
auth:
  authentication:
    enabled: false
    jwt:
      enabled: false
      # Enable JWT authentication
      # If the token is generated by a secret key, set the usingSecretKey as true.
      # If the token is generated by a private key, set the usingSecretKey as false.
      usingSecretKey: false
    openid:
      enabled: false
#     # https://pulsar.apache.org/docs/next/security-openid-connect/#enable-openid-connect-authentication-in-the-broker-and-proxy
      openIDAllowedTokenIssuers: []
      openIDAllowedAudiences: []
      openIDTokenIssuerTrustCertsFilePath:
      openIDRoleClaim:
      openIDAcceptedTimeLeewaySeconds: "0"
      openIDCacheSize: "5"
      openIDCacheRefreshAfterWriteSeconds: "64800"
      openIDCacheExpirationSeconds: "86400"
      openIDHttpConnectionTimeoutMillis: "10000"
      openIDHttpReadTimeoutMillis: "10000"
      openIDKeyIdCacheMissRefreshSeconds: "300"
      openIDRequireIssuersUseHttps: "true"
      openIDFallbackDiscoveryMode: "DISABLED"
```

### Verifying this change


frankjkelly:
Very cool - is there any presumption of what happens when the JWT token expires? Or if the client id / secret are rotated?

brunodomenici:
> Very cool - is there any presumption of what happens when the JWT token expires? Or if the client id / secret are rotated?

According with the [doc](https://pulsar.apache.org/docs/3.3.x/security-openid-connect/#openid-connect-authentication-flow), the client will handle the refresh:

> 8. When the token expires, the Pulsar Server challenges the client to re-authenticate with the Identity Provider and provide a new access token. If the client fails to re-authenticate, the Pulsar Server closes the connection.

I'm performing some tests. I let you know ;)

gulecroc:
Hi @lhotari, we would like to you OIDC support too, can someone review this PR ? Thank you

lhotari:
Closing and reopening to see if it triggers CI since the GitHub Actions workflow hasn't been run.

lhotari:
> Hi @lhotari, we would like to you OIDC support too, can someone review this PR ? Thank you

@gulecroc @brunodomenici Yes, OIDC support would be very useful. We'd also need a way to validate the configuration in GitHub Actions based CI. One possibility would be to have a lightweight OIDC provider get deployed in the kind cluster that the CI jobs use for validation. 

One of the best ways to debug CI is to enable GitHub Actions in your own fork and then open a PR to your own fork. When that is done, the GitHub Actions workflow will enable ssh access to the GitHub Actions Runner VMs for debugging failures.

charts/pulsar/values.yaml
```diff
@@ -273,12 +273,28 @@ tls:
 auth:
   authentication:
     enabled: false
-    provider: "jwt"
```
lhotari, line 276: Since many users don't read release notes, it's useful to add validation to prevent users from upgrading unless they remove deprecated config values.

Add a template called `_values_validation.tpl` and add validation with something like

```
{{- if (and .Values.auth.authentication.enabled (not (empty .Values.auth.authentication.provider))) }}
  {{- fail "ERROR: Setting auth.authentication.provider is no longer supported. For details, see the migration guide in README.md." }}
{{- end }}
```

lhotari:
> One possibility would be to have a lightweight OIDC provider get deployed in the kind cluster that the CI jobs use for validation.

Keycloak could be one choice. ClaudeAI generated this k8s yaml: https://gist.github.com/lhotari/8f483fe07f70e338192741867d0c7186 . There could be better ways, I just shared the ClaudeAI generated one. Something like that could be put in [the `.ci` directory](https://github.com/apache/pulsar-helm-chart/tree/master/.ci) and then having the CI script to simply use `kubectl` to install it to the kind cluster when the CI job is for testing OIDC.

charts/pulsar/values.yaml
lhotari: Please also add migration information to README.md

lhotari:
@brunodomenici @gulecroc  In addition to plain OIDC provider support, there would be a need to handle broker-to-broker and proxy-to-broker authentication. The foundations for this is described in the Pulsar docs at [Enable Custom OpenID Connect Integration with Kubernetes](https://pulsar.apache.org/docs/4.0.x/security-openid-connect/#enable-custom-openid-connect-integration-with-kubernetes). 
It's also possible to use static JWT tokens for broker-to-broker authentication while enabling OIDC auth, but the problem with that type of solution is that there's no current way to prevent using JWT tokens from outside of the cluster. A possible leak of the "superuser" token would be a problem. That's why it would be recommended to replace superuser tokens completely with Kubernetes OIDC provided tokens.

gulecroc:
Thank you @lhotari, I start working with @brunodomenici on your reviews

gulecroc:
Hi @lhotari, I configure [the keyclock bitnami chart as dependency](https://github.com/gulecroc/pulsar-helm-chart/blob/feat/openid/charts/pulsar/Chart.yaml) but I could not [download it during the CI](https://github.com/gulecroc/pulsar-helm-chart/actions/runs/14839455394/job/41658608573), maybe because of the OCI chart ? Bitnami switch only to OCI charts at the end of the last year

lhotari (approved):
LGTM, good work @brunodomenici

PR #517, opened 2024-08-14
Upgrade cert-manager to v1.12.13

lhotari:
- cert-manager 1.12 is a LTS release, EOL until May 2025


lhotari:
cert-manager 1.12 requires at least k8s 1.22 so this change is blocked until we lift the minimum k8s version requirement.

PR #523, opened 2024-08-22
Support using self generated certificates

duncaan:
Fixes #<xyz>

### Motivation

We get a SSL cert from another provider (not managed by cert-manager) and do not have a way to provide this to our proxies.

### Modifications

Added a variable that defaults to true and will remove the cert-manager certificate.

### Verifying this change


lhotari (approved):
LGTM! Thanks for the contribution

PR #525, opened 2024-09-03
feat: Support to customize broker podManagementPolicy from values.yaml

ludmanl:
Modification to make the podManagementPolicy on the broker adjustable, independent of the component function value, in cases where we do not want to use the built-in functions, but rather an extension like the FunctionMesh Operator.

### Motivation
podManagementPolicy should be customizable on the broker regardless of the value of the component function.

### Modifications
podManagementPolicy can be customized from the values file.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/broker-statefulset.yaml
+++ b/charts/pulsar/templates/broker-statefulset.yaml
@@ -49,7 +49,7 @@ spec:
   podManagementPolicy: {{ $stsObj.spec.podManagementPolicy }}
   {{- else }}
   {{- if not .Values.components.functions }}
-  podManagementPolicy: Parallel
+  podManagementPolicy: {{ .Values.broker.podManagementPolicy }}
   {{- else }}
   podManagementPolicy: OrderedReady
   {{- end }}
--- a/charts/pulsar/templates/broker-statefulset.yaml
+++ b/charts/pulsar/templates/broker-statefulset.yaml
@@ -48,8 +48,10 @@ spec:
   {{- if $stsObj }}
   podManagementPolicy: {{ $stsObj.spec.podManagementPolicy }}
   {{- else }}
-  {{- if not .Values.components.functions }}
+  {{- if .Values.broker.podManagementPolicy }}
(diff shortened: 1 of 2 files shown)


charts/pulsar/templates/broker-statefulset.yaml
```diff
@@ -49,7 +49,7 @@ spec:
   podManagementPolicy: {{ $stsObj.spec.podManagementPolicy }}
   {{- else }}
   {{- if not .Values.components.functions }}
-  podManagementPolicy: Parallel
+  podManagementPolicy: {{ .Values.broker.podManagementPolicy }}
   {{- else }}
   podManagementPolicy: OrderedReady
   {{- end }}
```
lhotari, line 55: I think that we'd need to take this existing logic into account. The default value in the values.yaml should be absent (nil). When the value is absent, the former logic should be used. If it's set, it would get applied.

charts/pulsar/values.yaml
```diff
@@ -806,6 +806,7 @@ broker:
     maxReplicas: 3
     metrics: ~
     behavior: ~
+  podManagementPolicy: Parallel
```
lhotari, line 809: simply provide the key without the value, so that it's absent by default.
Please add a short documentation to the key. It's worth mentioning that the value cannot be modified without manual operations after a cluster has been deployed.

lhotari (changes requested):

charts/pulsar/values.yaml
```diff
@@ -806,6 +806,8 @@ broker:
     maxReplicas: 3
     metrics: ~
     behavior: ~
+  # The podManagementPolicy cannot be changed on an existing broker. If you want to make this change, you will need to manually recreate the StatefulSet.
```
lhotari, line 809:
```suggestion
  # The podManagementPolicy cannot be modified for an existing deployment. If you need to change this value, you will need to manually delete the existing broker StatefulSet and then redeploy the chart.
```

lhotari (approved):

PR #535, opened 2024-10-03
feat: add extraVolumes and Mounts for pulsar-manager

### LINKED ISSUE #534: add extraVolumes and extraVolumeMounts support to the pulsar_manager
Is your feature request related to a problem? Please describe.
We need to mount the JWT secret in the pulsar manager. We don't use the `.Values.auth` configuration because we only want to use the JWT on the pulsar manager. The `Values.auth` configures the whole system in JWT and that's not what we want because we're using another plugin for authentication.

Describe the solution you'd like
add extraVolumes and extraVolumeMounts support to the pulsar_manager values in helm chart

lenglet-k:
Fixes #534


lhotari (approved):
LGTM, thanks for the contribution @lenglet-k

PR #540, opened 2024-10-15
fix: broker extraEnv variable

### LINKED ISSUE #539: Broker environnement variable cannot be set
**Describe the bug**
If we want add environment variable with extreEnvs object in broker is not possible because `env:` is declared twice in yaml manifest.

Here:  https://github.com/apache/pulsar-helm-chart/blob/master/charts/pulsar/templates/broker-statefulset.yaml#L266
and Here: https://github.com/apache/pulsar-helm-chart/blob/master/charts/pulsar/templates/broker-statefulset.yaml#L298

**To Reproduce**
1. Set environment variable in broker.extreEnvs
2. Deploy helm release
3. See your pod deployed by broker statefulset without your environment variable

**Expected behavior**
Set environment variable for broker

lenglet-k:
Fixes #539

### Motivation

fix `env` object declared twice ni broker.yaml

QUESTION: It's possible to rename extreEnvs to extraEnvs  ?

### Modifications

*Use only one env object in broker*

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/broker-statefulset.yaml
+++ b/charts/pulsar/templates/broker-statefulset.yaml
@@ -316,8 +316,8 @@ spec:
                 name: {{ .Values.broker.storageOffload.secret }}
                 key: AZURE_STORAGE_ACCESS_KEY
           {{- end }}
-        {{- if .Values.broker.extreEnvs }}
-{{- toYaml .Values.broker.extreEnvs | nindent 10 }}
+        {{- if .Values.broker.extraEnvs }}
+{{- toYaml .Values.broker.extraEnvs | nindent 10 }}
         {{- end }}
       volumes:
       {{- if .Values.broker.extraVolumes }}
--- a/charts/pulsar/templates/broker-statefulset.yaml
+++ b/charts/pulsar/templates/broker-statefulset.yaml
@@ -262,10 +262,6 @@ spec:
         - name: "{{ .Values.tlsPrefix }}pulsarssl"
           containerPort: {{ .Values.broker.ports.pulsarssl }}
         {{- end }}
-{{- if .Values.broker.extreEnvs }}
-        env:
-{{ toYaml .Values.broker.extreEnvs | indent 8 }}
-{{- end }}
         envFrom:
         - configMapRef:
             name: "{{ template "pulsar.fullname" . }}-{{ .Values.broker.component }}"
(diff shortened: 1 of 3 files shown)


lhotari:
> QUESTION: It's possible to rename extreEnvs to extraEnvs ?

Makes sense to fix the typo.

lenglet-k:
> > QUESTION: It's possible to rename extreEnvs to extraEnvs ?
> 
> Makes sense to fix the typo.

it can cause breaking change if I rename, can I avoid breaking change by keeping the old value in the if ( or .values.broker.extraEnvs .value.broker.extreEnvs )? @lhotari

lenglet-k:
> > > > QUESTION: It's possible to rename extreEnvs to extraEnvs ?
> > > 
> > > 
> > > Makes sense to fix the typo.
> > 
> > 
> > it can cause breaking change if I rename, can I avoid breaking change by keeping the old value in the if ( or .values.broker.extraEnvs .value.broker.extreEnvs )? @lhotari
> 
> It's fine. I'll add a note to the next release notes about the change. I'll be bumping the version to 3.7.0 with the next release to indicate that there could be minor breaking changes.

Ok so, i rename only to extraEnvs. And an other question, what's the cycle of release (because we wait 3.6.0 ^^' )?

lhotari:
> Ok so, i rename only to extraEnvs. And an other question, what's the cycle of release (because we wait 3.6.0 ^^' )?

3.6.0 was released in the last hour: https://lists.apache.org/thread/2xg6czzp6xqr829xjvzzx1tvfwzjov6c
3.7.0 will be released [after Pulsar 4.0.0 which is currently in voting phase](https://lists.apache.org/thread/jqnvrh45tly9yzyvgcq33v7tsqv12ng3). ETA for 3.7.0 with Pulsar 4.0.0 by default is Oct 28th.
btw. Please test Pulsar 4.0.0-candidate-1 to find any possible critical regressions.

lhotari:
> we use argocd to deploy the charts and it doesn't seem to find your candidate versions. I need to check this problem more carefully

Docker images for 4.0.0-candidate-1 (https://lists.apache.org/thread/jqnvrh45tly9yzyvgcq33v7tsqv12ng3):
docker pull lhotari/pulsar:4.0.0-92448d5
https://hub.docker.com/layers/lhotari/pulsar/4.0.0-92448d5/images/sha256-285b28c50308f2a452a58b6a57529fbcf41517f46c2cae2cd9482659f9491a97?context=explore
docker pull lhotari/pulsar-all:4.0.0-92448d5
https://hub.docker.com/layers/lhotari/pulsar-all/4.0.0-92448d5/images/sha256-4111ea92fde9b74f18165c42ebe9039447d01847bec9c71796a31776cdbea9f8?context=explore

lhotari:
> we use argocd to deploy the charts and it doesn't seem to find your candidate versions. I need to check this problem more carefully

in some cases, it might be necessary to prefix the docker image name with `docker.io/` for example `docker.io/lhotari/pulsar:4.0.0-92448d5` / `docker.io/lhotari/pulsar-all:4.0.0-92448d5`.

lhotari (approved):
LGTM, thanks for contributing

PR #548, opened 2024-11-11
Correct pulsar proxy prometheus.io/port annotation

### LINKED ISSUE #547: Pulsar proxy prometheus.io/port annotation uses the service port value instead of the containerPort value
**Describe the bug**
The prometheus.io/port annotation for the proxy statefulset uses the port value for the service instead of the value for the container.  If those two things are not the same (which they aren't by default), prometheus scrapes will fail.

**To Reproduce**
1. deploy pulsar proxy using the helm chart and default values in a kubernetes cluster where prometheus is configured to scrape pods based on annotations
2. Notice prometheus is trying to scrape port 80, when the pods are actually listening on port 8080

**Expected behavior**
Prometheus scrap annotations should contain the port the pods are listening on, allowing them to be scraped.

This is a simple fix.  I'll be submitting a PR to correct it.

doug-ba:
Fixes #547 

### Motivation
Being able to scrape metrics is good.  The existing version of the chart can be worked around by setting both the service and containerPort http ports to the same value, but the default values file has them different, and it's a confusing gotcha.

### Modifications
Corrected the annotation to use the value used to set the containerPort instead of the value used to set the service port.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/proxy-statefulset.yaml
+++ b/charts/pulsar/templates/proxy-statefulset.yaml
@@ -45,7 +45,7 @@ spec:
         component: {{ .Values.proxy.component }}
       annotations:
         prometheus.io/scrape: "true"
-        prometheus.io/port: "{{ .Values.proxy.ports.http }}"
+        prometheus.io/port: "{{ .Values.proxy.ports.containerPorts.http }}"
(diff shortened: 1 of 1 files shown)


lhotari (approved):
LGTM, thanks for the contribution @doug-ba

PR #550, opened 2024-11-18
Add support for admin port on ZooKeeper

### LINKED ISSUE #549: Unable to access ZooKeeper admin port
We need to be able to access the ZooKeeper admin endpoints to perform snapshots, and administrative actions. The Helm chart currently does not expose this port (9990) [on the service](https://github.com/apache/pulsar-helm-chart/blob/master/charts/pulsar/templates/zookeeper-service.yaml), or [statefulset specification](https://github.com/apache/pulsar-helm-chart/blob/master/charts/pulsar/templates/zookeeper-statefulset.yaml#L134).

Allow the user to specify the admin port number in the Helm values to expose the admin port on the service.

Considered changing the admin port number to one that is not commonly used, but we're using all other ports.

**Additional context**
This is required so we can get the zxid of each ZooKeeper instances, and then perform a snapshot on the highest instance.

liamgib:
Fixes #549

### Motivation

We need to be able to access the ZooKeeper admin endpoint to perform automated snapshots to an external system. 

### Modifications

I have added the admin port to the list of Zookeeper ports, and made the relevant changes on the zookeeper-service.yaml & zookeeper-statefulset.yaml to use the admin port.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/zookeeper-service.yaml
+++ b/charts/pulsar/templates/zookeeper-service.yaml
@@ -40,6 +40,8 @@ spec:
       port: {{ .Values.zookeeper.ports.leaderElection }}
     - name: "{{ .Values.tcpPrefix }}client"
       port: {{ .Values.zookeeper.ports.client }}
+    - name: admin
+      port: {{ .Values.zookeeper.ports.admin }}
     {{- if and .Values.tls.enabled .Values.tls.zookeeper.enabled }}
     - name: "{{ .Values.tlsPrefix }}client-tls"
       port: {{ .Values.zookeeper.ports.clientTls }}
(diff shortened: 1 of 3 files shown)


lhotari (commented):
@liamgib Thanks for the contribution! Are there security implications of exposing the admin port by default?

liamgib:
Good point @lhotari, there is some security implications. Especially for existing clusters, where they might not expecting the admin port being exposed.

I'll update the PR to make the field optional, and removed by default.

charts/pulsar/templates/zookeeper-service.yaml
```diff
@@ -40,6 +40,10 @@ spec:
       port: {{ .Values.zookeeper.ports.leaderElection }}
     - name: "{{ .Values.tcpPrefix }}client"
       port: {{ .Values.zookeeper.ports.client }}
+    {{- if .Values.zookeeper.ports.admin }}
```
liamgib, line 43: Happy to move this to a boolean style if preferred

lhotari (approved):
LGTM

PR #558, opened 2024-12-09
Fix helm chart to allow configurable ingress pathType

rawmind0:
### Motivation

The ingress configuration for pulsar_manager and for proxy components are failing on some circumstances due to `pathType` is hardcoded.

### Modifications

The proposed modifications is that `pathType` value may be modified from values.yaml file. The change has set the `pathType` default value to the same that was harcoded `pathType: ImplementationSpecific` so the change is backward compatible.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/proxy-ingress.yaml
+++ b/charts/pulsar/templates/proxy-ingress.yaml
@@ -59,7 +59,7 @@ spec:
               servicePort: {{ .Values.proxy.ports.http }}
               {{- end }}
             {{- else }}
-            pathType: ImplementationSpecific
+            pathType: {{ .Values.proxy.ingress.pathType }}
             backend:
               service:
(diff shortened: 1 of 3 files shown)


lhotari (approved):
LGTM, thanks for contributing @rawmind0

jleni:
any chance this could be included in a patch release soon?

lhotari:
> any chance this could be included in a patch release soon? 3.8.1 maybe?

@jleni In Apache Pulsar projects we don't have an automated release pipeline due to [Apache Software Foundation (ASF) release policies](https://www.apache.org/legal/release-policy.html). Each release requires voting to happen with 3 binding votes from Apache Pulsar PMC members. That's why releases aren't frequent and they are relatively slow. For the Pulsar Helm chart, I'd recommend forking the repository and using the fork for internal deployments so that you aren't dependent on official releases.

PR #561, opened 2024-12-13
feat(tls): support ca type issuer and v1alpha* version cert-manager api

ericsyh:
Fixes #<xyz>

### Motivation

The current chart only supports enable tls with self-signed issuer which is not possible for production users. This PR supports [ca type issuser](https://cert-manager.io/docs/configuration/ca/) which allows users to bound a root certificate (in `Secret`) to issue required tls certificates for pulsar components. 

### Modifications

- support the ca type issuer in the `certs.type` and add the `certs.issuers.ca.secretName` field for the root certificate reference. 
- support cert-manager v1alpha[1-3] API versions. 

### Verifying this change

- [x] Make sure that the change passes the CI checks.

PR #565, opened 2025-01-22
add template for ca issuer name and secret name

gulecroc:
### Motivation

To test geo-replication, I would like to install the pulsar chart 3 times :
* 1 for global zookeeper
* 2 for 2 pulsar clusters

I also would like to use TLS. The global zookeeper install will create the CA issuer and certificate, and the 2 pulsar clusters will use it to create their certificates.

### Modifications

* add template to override CA issuer name and certificate secret name

### Verifying this change


charts/pulsar/templates/_certs.tpl
```diff
@@ -0,0 +1,40 @@
...
+{{/*
+Define the pulsar certs ca issuer secret name
+*/}}
+{{- define "pulsar.certs.issuers.ca.secretName" -}}
+{{- if .Values.certs.issuers.ca.secretName -}}
+{{- .Values.certs.issuers.ca.secretName -}}
+{{- else -}}
+{{ printf "%s-%s" .Release.Name .Values.tls.ca_suffix }}
+{{- end -}}
+{{- end -}}
```
lhotari, line 40: This is a great improvement.
One additional request would be to handle both .Values.certs.internal_issuer.type cases in these templates, for both "selfsigning" and "ca". That would make this PR more consistent. (#561 recently added/improved "ca" handling).

lhotari (commented):
Thanks for the PR, @gulecroc . Please remove the geo-replication changes out of this PR since that will simplify handling of this PR so that we first get the TLS templating changes made.
We can handle the geo-replication examples in a separate PR.

I really appreciate your effort in making a Geo-replication example available directly in the Pulsar helm chart.
One minor detail is that global Zookeeper isn't required in Pulsar to use geo-replication. For completeness of the examples, it would be useful to then have examples of both cases. The Pulsar documentation doesn't clearly explain the different ways Pulsar geo-replication can be configured.

charts/pulsar/templates/proxy-statefulset.yaml
```diff
@@ -300,7 +300,7 @@ spec:
         - name: ca
           secret:
             {{- if eq .Values.certs.internal_issuer.type "selfsigning" }}
```
lhotari, line 302: It looks like `secretName: "{{ template "pulsar.certs.issuers.ca.secretName" . }}"` should be used also here for consistency.

charts/pulsar/templates/toolset-statefulset.yaml
```diff
@@ -126,7 +126,7 @@ spec:
       - name: proxy-ca
         secret:
           {{- if eq .Values.certs.internal_issuer.type "selfsigning" }}
-          secretName: "{{ .Release.Name }}-{{ .Values.tls.ca_suffix }}"
+          secretName: "{{ template "pulsar.certs.issuers.ca.secretName" . }}"
           {{- end }}
           {{- if eq .Values.certs.internal_issuer.type "ca" }}
           secretName: "{{ .Values.certs.issuers.ca.secretName }}"
```
lhotari, line 132: It looks like `secretName: "{{ template "pulsar.certs.issuers.ca.secretName" . }}"` should be used also here for consistency.

charts/pulsar/templates/zookeeper-statefulset.yaml
```diff
@@ -254,7 +254,7 @@ spec:
       - name: ca
         secret:
           {{- if eq .Values.certs.internal_issuer.type "selfsigning" }}
-          secretName: "{{ .Release.Name }}-{{ .Values.tls.ca_suffix }}"
+          secretName: "{{ template "pulsar.certs.issuers.ca.secretName" . }}"
           {{- end }}
           {{- if eq .Values.certs.internal_issuer.type "ca" }}
           secretName: "{{ .Values.certs.issuers.ca.secretName }}"
```
lhotari, line 260: It looks like `secretName: "{{ template "pulsar.certs.issuers.ca.secretName" . }}"` should be used also here for consistency.

charts/pulsar/templates/proxy-statefulset.yaml
lhotari: pushed a commit to address this.

Issue #567, opened 2025-02-01
Increase defaults for ensemble size, write quorum, and ack quorum to 2


pdolif:
This issue is based on this discussion: https://github.com/apache/pulsar/discussions/23528

Currently, the default values for the ensemble size, write quorum, and ack quorum are 1.

https://github.com/apache/pulsar-helm-chart/blob/e55405cbe2819647f1a642a850364f17544603b0/charts/pulsar/values.yaml#L1019-L1021

As mentioned by @lhotari "2 would be a more sensible default". Can these defaults simply be set to 2 or is there something else to consider?

illegalnumbers:
@lhotari I would be willing to make this change tomorrow if you are ok with it.

lhotari:
Addressed by #570

PR #570, opened 2025-02-08
Increase defaults for ensemble size, write quorum, and ack quorum to 2

pdolif:
#567

### Motivation

Currently, the default values for the ensemble size, write quorum, and ack quorum are 1.
2 is a more suitable default.


lhotari (approved):
LGTM. Thanks for the contribution, @pdolif!

PR #592, opened 2025-03-08
Migrate from kube-prometheus-metrics to victoria-metrics-k8s-stack

lhotari:
### Motivation

kube-prometheus-stack has been replaced by victoria-metrics-k8s-stack for Pulsar Helm Chart version 4.0.0 due to
Prometheus incompatibility with Pulsar metrics since Pulsar 2.11.0 . Pulsar exposes metrics in a format that is partially OpenMetrics 1.0.0 text format, but isn't fully compatible. Prometheus doesn't provide proper support for OpenMetrics 1.0.0 text format, even in Prometheus version 3.2.1 where it was extensively tested before making the switch to Victoria MetricsVictoria Metrics is Apache 2.0 Licensed OSS and it's fully compatible with Prometheus.

### Modifications

- replace kube-prometheus-stack with [victoria-metrics-k8s-stack ](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-k8s-stack/) which is Apache 2.0 Licensed OSS and it's fully compatible with Prometheus.
- The Pulsar Helm Chart has pre-installed Grafana dashboards for Pulsar from various Apache 2.0 Licensed sources. These dashboards are outdated, and there's a separate project https://github.com/lhotari/pulsar-grafana-dashboards where the dashboards will be cleaned up and made fully compatible with Pulsar 4.0+ and Victoria Metrics.
  - In Pulsar, the Prometheus client_java library was updated to 0.15.0 with https://github.com/apache/pulsar/pull/13785 in Pulsar 2.11.0 and this broke most of the metrics since the output is not Prometheus Text 0.0.4 format any more. 
  - Since Victoria Metrics has better support for handling scraped metrics, it should be possible to get the metric names working again with Victoria Metrics configuration. In Prometheus, this isn't possible at the moment. There are some changes going on in the Prometheus project, to address it, but since Victoria Metrics is more feature complete and runs with less resources, it's a better choice in getting the dashboards usable again.

### Additional context

- Another PR will contain changes to migrate from using Prometheus Operator CRDs to use Victoria Metrics CRDs (PodMonitor -> VMPodScrape)

### DIFF
--- a/.ci/chart_test.sh
+++ b/.ci/chart_test.sh
@@ -60,7 +60,11 @@ if [[ "$UPGRADE_FROM_VERSION" != "" ]]; then
     ALLOW_LOADBALANCERS="true"
     # install older version of pulsar chart
     PULSAR_CHART_VERSION="$UPGRADE_FROM_VERSION"
-    ci::install_pulsar_chart install ${PULSAR_HOME}/.ci/values-common.yaml ${PULSAR_HOME}/${VALUES_FILE} "${extra_opts[@]}"
+
+    # Install Prometheus Operator CRDs using the upgrade script since kube-prometheus-stack is now disabled before the upgrade
+    ${PULSAR_HOME}/scripts/kube-prometheus-stack/upgrade_prometheus_operator_crds.sh
+
+    ci::install_pulsar_chart install ${PULSAR_HOME}/.ci/values-common.yaml ${PULSAR_HOME}/${VALUES_FILE} --set kube-prometheus-stack.enabled=false "${extra_opts[@]}"
     install_type="upgrade"
     echo "Wait 10 seconds"
     sleep 10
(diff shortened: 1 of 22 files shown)


mrtworo:
@lhotari could you share some details on impact of staying with `kube-prometheus-stack`? Which metrics are broken? I couldn't find it here or in `pulsar` repo, sorry if it was already published.

lhotari:
> @lhotari could you share some details on impact of staying with `kube-prometheus-stack`? Which metrics are broken? I couldn't find it here or in `pulsar` repo, sorry if it was already published.

@mrtworo I had limited time to spend on it and I seemed to have ran into some Prometheus bugs. one of them contained similar error messages as https://github.com/prometheus/prometheus/issues/16160 . To be actually sure about the reasons, it would require time and effort that I don't have. With victoria-metrics-k8s-stack, I didn't run into problems and I was able to get the dashboards into a usable state.
One challenge was https://prometheus.io/docs/prometheus/latest/migration/#scrape-protocols which I was able to resolve with `fallback_scrape_protocol`. There's only a single metric that uses a histogram in Pulsar metrics (the metadata one) and Prometheus wasn't able to understand that. That was a minor issue. With VictoriaMetrics, the Pulsar metrics get ingested without issues. My observations could always be incorrect. Switching from Prometheus to VictoriaMetrics seems to be useful as a default. If someone would like to continue using kube-prometheus-stack, they could always customize their chart usage. There's simply not sufficient interest and contributors in this project to keep too many options around.

Issue #602, opened 2025-04-29
Broker initialization error when using global Zookeeper


trynocoding:
**Describe the bug**
```
2025-04-29T05:42:01,615+0000 [main] INFO  org.apache.zookeeper.ClientCnxn - zookeeper.request.timeout value is 0. feature enabled=false
2025-04-29T05:42:01,624+0000 [main-SendThread(pulsar-cs-zookeeper.pulsar-cs.svc.cluster.local:2181)] INFO  org.apache.zookeeper.ClientCnxn - Opening socket connection to server pulsar-cs-zookeeper.pulsar-cs.svc.cluster.local/10.0.1.61:2181.
2025-04-29T05:42:01,625+0000 [main-SendThread(pulsar-cs-zookeeper.pulsar-cs.svc.cluster.local:2181)] INFO  org.apache.zookeeper.ClientCnxn - SASL config status: Will not attempt to authenticate using SASL (unknown error)
2025-04-29T05:42:01,633+0000 [main-SendThread(pulsar-cs-zookeeper.pulsar-cs.svc.cluster.local:2181)] INFO  org.apache.zookeeper.ClientCnxn - Socket connection established, initiating session, client: /10.0.0.144:43116, server: pulsar-cs-zookeeper.pulsar-cs.svc.cluster.local/10.0.1.61:2181
2025-04-29T05:42:01,645+0000 [main-SendThread(pulsar-cs-zookeeper.pulsar-cs.svc.cluster.local:2181)] INFO  org.apache.zookeeper.ClientCnxn - Session establishment complete on server pulsar-cs-zookeeper.pulsar-cs.svc.cluster.local/10.0.1.61:2181, session id = 0x200757815ff0028, negotiated timeout = 30000

WATCHER::

WatchedEvent state:SyncConnected type:None path:null zxid: -1
Node does not exist: /admin/clusters/cluster3
2025-04-29T05:42:01,657+0000 [main] ERROR org.apache.zookeeper.util.ServiceUtils - Exiting JVM with code 1
pulsar cluster cluster3 isn't initialized yet ... check in 3 seconds ...
```

**To Reproduce**
1. helm install pulsar-cs apache/pulsar -f config-store-values.yaml
2. helm install cluster3 apache/pulsar -f cluster3-values.yaml --namespace pulsar-cluster3 --create-namespace --set affinity.anti_affinity=false --set kube-prometheus-stack.enabled=false --set clusterName=cluster3
 ```
[...]

PR #603, opened 2025-04-29
Fix broker initialization error when using global Zookeeper (#602)

trynocoding:
Fixes #<xyz>

### Motivation

*Explain here the context, and why you're making that change. What is the problem you're trying to solve.*

### Modifications

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/proxy-statefulset.yaml
+++ b/charts/pulsar/templates/proxy-statefulset.yaml
@@ -125,10 +125,14 @@ spec:
             {{- if $zk:=.Values.pulsar_metadata.userProvidedZookeepers }}
             until timeout 15 bin/pulsar zookeeper-shell -server {{ $zk }} ls {{ or .Values.metadataPrefix "/" }}; do
               echo "user provided zookeepers {{ $zk }} are unreachable... check in 3 seconds ..." && sleep 3;
+            done; 
+            {{- else if .Values.pulsar_metadata.configurationStore }}
+            until timeout 15 bin/pulsar zookeeper-shell -server {{ template "pulsar.configurationStore.service" . }} get {{ .Values.pulsar_metadata.configurationStoreMetadataPrefix }}/admin/clusters/{{ template "pulsar.cluster.name" . }}; do
+              echo "pulsar cluster {{ template "pulsar.cluster.name" . }} isn't initialized yet ... check in 3 seconds ..." && sleep 3;
             done;
-            {{ else }}
-            until timeout 15 bin/pulsar zookeeper-shell -server {{ template "pulsar.configurationStore.service" . }} get {{ .Values.metadataPrefix }}/admin/clusters/{{ template "pulsar.cluster.name" . }}; do
-              sleep 3;
+            {{- else }}
+            until timeout 15 bin/pulsar zookeeper-shell -server {{ template "pulsar.zookeeper.connect" . }} get {{ .Values.metadataPrefix }}/admin/clusters/{{ template "pulsar.cluster.name" . }}; do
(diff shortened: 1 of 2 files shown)


charts/pulsar/templates/proxy-statefulset.yaml
```diff
@@ -127,7 +127,7 @@ spec:
               echo "user provided zookeepers {{ $zk }} are unreachable... check in 3 seconds ..." && sleep 3;
             done;
             {{ else }}
-            until timeout 15 bin/pulsar zookeeper-shell -server {{ template "pulsar.configurationStore.service" . }} get {{ .Values.metadataPrefix }}/admin/clusters/{{ template "pulsar.cluster.name" . }}; do
+            until timeout 15 bin/pulsar zookeeper-shell -server {{ template "pulsar.configurationStore.service" . }} get {{ .Values.pulsar_metadata.configurationStoreMetadataPrefix }}/admin/clusters/{{ template "pulsar.cluster.name" . }}; do
```
lhotari, line 130: the `{{- if .Values.pulsar_metadata.configurationStore }}` condition is missing in this template. Please check this change.

lhotari (approved):
LGTM

lhotari:
@trynocoding There's some failures with TLS tests.

trynocoding:
Closing and reopening to trigger a fresh test run to see if the problems persist. @trynocoding Please investigate the TLS test issues if they persist.

It appears that the proxy init container is missing TLS configuration.

charts/pulsar/templates/proxy-statefulset.yaml
```diff
@@ -125,10 +125,14 @@ spec:
...
+            done; 
+            {{- else if .Values.pulsar_metadata.configurationStore }}
+            until timeout 15 bin/pulsar zookeeper-shell -server {{ template "pulsar.configurationStore.service" . }} get {{ .Values.pulsar_metadata.configurationStoreMetadataPrefix }}/admin/clusters/{{ template "pulsar.cluster.name" . }}; do
+              echo "pulsar cluster {{ template "pulsar.cluster.name" . }} isn't initialized yet ... check in 3 seconds ..." && sleep 3;
             done;
-            {{ else }}
-            until timeout 15 bin/pulsar zookeeper-shell -server {{ template "pulsar.configurationStore.service" . }} get {{ .Values.metadataPrefix }}/admin/clusters/{{ template "pulsar.cluster.name" . }}; do
-              sleep 3;
+            {{- else }}
+            until timeout 15 bin/pulsar zookeeper-shell -server {{ template "pulsar.zookeeper.service" . }} get {{ .Values.metadataPrefix }}/admin/clusters/{{ template "pulsar.cluster.name" . }}; do
```
lhotari, line 134: In other locations, `pulsar.zookeeper.connect` template is used. I guess it takes user provided zk servers into account.
```suggestion
            until timeout 15 bin/pulsar zookeeper-shell -server {{ template "pulsar.zookeeper.connect" . }} get {{ .Values.metadataPrefix }}/admin/clusters/{{ template "pulsar.cluster.name" . }}; do
```

trynocoding:
The previous TLS scenario failed because pulsar.zookeeper.connect was used. This PR previously used pulsar.configurationStore.service (which is essentially equivalent to pulsar.zookeeper.connect), so the TLS issue that exists was never exposed. The root cause is that wait-zookeeper-ready should also be configured with TLS certificate information, which is currently missing. The reason for continuing to use pulsar.zookeeper.connect is to maintain the original logic, and it's better to address the TLS fix in a separate PR.

lhotari (approved):

PR #605, opened 2025-05-02
oxia components podmonitor match labels

anosulchik:
Fixes mismatch between podmonitor for oxia (coordinator and server) and labels for the pods.

### Motivation

This is a bug that prevents podmonitor to work properly - it can't match pods so metrics won't be ingested. For example, for oxia-server pods it's set to `component: oxia-server` but in podMonitor it's `kubernetes.io/component: oxia-server`.

### Modifications

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/oxia-coordinator-podmonitor.yaml
+++ b/charts/pulsar/templates/oxia-coordinator-podmonitor.yaml
@@ -19,5 +19,5 @@
 
 # deploy oxia-coordinator PodMonitor only when `$.Values.oxia.coordinator.podMonitor.enabled` is true
 {{- if and $.Values.components.oxia $.Values.oxia.coordinator.podMonitor.enabled }}
-{{- include "pulsar.podMonitor" (list . "oxia.coordinator" "app.kubernetes.io/component: oxia-coordinator" "metrics") }}
+{{- include "pulsar.podMonitor" (list . "oxia.server" (printf "component: %s-coordinator" .Values.oxia.component)) }}
 {{- end }}
--- a/charts/pulsar/templates/oxia-coordinator-podmonitor.yaml
+++ b/charts/pulsar/templates/oxia-coordinator-podmonitor.yaml
@@ -19,5 +19,5 @@
 
 # deploy oxia-coordinator PodMonitor only when `$.Values.oxia.coordinator.podMonitor.enabled` is true
 {{- if and $.Values.components.oxia $.Values.oxia.coordinator.podMonitor.enabled }}
-{{- include "pulsar.podMonitor" (list . "oxia.coordinator" (printf "component: %s-coordinator" .Values.oxia.component)) }}
+{{- include "pulsar.podMonitor" (list . "oxia.coordinator" (printf "component: %s-coordinator" .Values.oxia.component) "metrics") }}
 {{- end }}
--- a/charts/pulsar/templates/oxia-coordinator-podmonitor.yaml
+++ b/charts/pulsar/templates/oxia-coordinator-podmonitor.yaml
@@ -19,5 +19,5 @@
 
 # deploy oxia-coordinator PodMonitor only when `$.Values.oxia.coordinator.podMonitor.enabled` is true
 {{- if and $.Values.components.oxia $.Values.oxia.coordinator.podMonitor.enabled }}
-{{- include "pulsar.podMonitor" (list . "oxia.server" (printf "component: %s-coordinator" .Values.oxia.component)) }}
+{{- include "pulsar.podMonitor" (list . "oxia.coordinator" (printf "component: %s-coordinator" .Values.oxia.component)) }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


charts/pulsar/templates/oxia-coordinator-podmonitor.yaml
```diff
@@ -19,5 +19,5 @@
 
 # deploy oxia-coordinator PodMonitor only when `$.Values.oxia.coordinator.podMonitor.enabled` is true
 {{- if and $.Values.components.oxia $.Values.oxia.coordinator.podMonitor.enabled }}
-{{- include "pulsar.podMonitor" (list . "oxia.coordinator" "app.kubernetes.io/component: oxia-coordinator" "metrics") }}
-{{- end }}
\ No newline at end of file
+{{- include "pulsar.podMonitor" (list . "oxia.server" (printf "component: %s-coordinator" .Values.oxia.component)) }}
```
lhotari, line 22: I guess `oxia.server` should be replaced with `oxia.coordinator` here:
```suggestion
{{- include "pulsar.podMonitor" (list . "oxia.coordinator" (printf "component: %s-coordinator" .Values.oxia.component)) }}
```

lhotari (commented):
Thanks for the contribution, @anosulchik! One small detail to check.

charts/pulsar/templates/oxia-coordinator-podmonitor.yaml
anosulchik: Thanks, this was a typo - it's fixed now.

charts/pulsar/templates/oxia-coordinator-podmonitor.yaml
```diff
@@ -19,5 +19,5 @@
 
 # deploy oxia-coordinator PodMonitor only when `$.Values.oxia.coordinator.podMonitor.enabled` is true
 {{- if and $.Values.components.oxia $.Values.oxia.coordinator.podMonitor.enabled }}
-{{- include "pulsar.podMonitor" (list . "oxia.coordinator" "app.kubernetes.io/component: oxia-coordinator" "metrics") }}
```
lhotari, line 22: one more detail is that "metrics" should be passed as the 4th parameter. That's the port name to be used.
https://github.com/apache/pulsar-helm-chart/blob/7833e51c282527e6b0c229f9bb5bae043e7a54f8/charts/pulsar/templates/_monitor.tpl#L25-L27
https://github.com/apache/pulsar-helm-chart/blob/7833e51c282527e6b0c229f9bb5bae043e7a54f8/charts/pulsar/values.yaml#L514
https://github.com/apache/pulsar-helm-chart/blob/7833e51c282527e6b0c229f9bb5bae043e7a54f8/charts/pulsar/templates/oxia-coordinator-deployment.yaml#L63-L67

charts/pulsar/templates/oxia-server-podmonitor.yaml
```diff
@@ -19,5 +19,5 @@
 
 # deploy oxia-server PodMonitor only when `$.Values.oxia.server.podMonitor.enabled` is true
 {{- if and $.Values.components.oxia $.Values.oxia.server.podMonitor.enabled }}
-{{- include "pulsar.podMonitor" (list . "oxia.server" "app.kubernetes.io/component: oxia-server" "metrics") }}
```
lhotari, line 22: one more detail is that "metrics" should be passed as the 4th parameter. That's the port name to be used.

lhotari (commented):
Please check passing of the 4th argument where it was "metrics" before.

charts/pulsar/templates/oxia-coordinator-podmonitor.yaml
anosulchik: @lhotari 

It doesn't seem to be working:

```
 # deploy oxia-server PodMonitor only when `$.Values.oxia.server.podMonitor.enabled` is true
 {{- if and $.Values.components.oxia $.Values.oxia.server.podMonitor.enabled }}
{{- include "pulsar.podMonitor" (list . "oxia.server" (printf "component: %s-server" .Values.oxia.component)) "metrics" }}
 {{- end }}
```

results in error:

```
Error: template: pulsar/templates/oxia-server-podmonitor.yaml:22:4: executing "pulsar/templates/oxia-server-podmonitor.yaml" at <include>: wrong number of args for include: want 2 got 3
```

I also see that port param is not included in other podmonitors:

```
# deploy bookkeeper PodMonitor only when `$.Values.bookkeeper.podMonitor.enabled` is true
{{- if $.Values.bookkeeper.podMonitor.enabled }}
{{- include "pulsar.podMonitor" (list . "bookkeeper" (printf "component: %s" .Values.bookkeeper.component)) }}
{{- end }}
```

Also, it looks like 4th param is supposed to be optional - it doesn't work as expected since adding "metrics" above leads to an error.

```
# templates/_monitor.tpl
...
{{- define "pulsar.podMonitor" -}}
{{- $root := index . 0 }}
{{- $component := index . 1 }}
{{- $matchLabel := index . 2 }}
{{- $portName := "http" }}
{{- if gt (len .) 3 }}
{{- $portName = index . 3 }}
{{- end }}

...
```

anosulchik: Disregard this, it works with proper formatting.

charts/pulsar/templates/oxia-server-podmonitor.yaml
anosulchik: Thanks. It's added now.

lhotari (approved):
LGTM

PR #606, opened 2025-05-05
configurable oxia coordinator configmap and entrypoint

anosulchik:
It enables possibility to configure oxia coordinator and its configmap externally.

### Motivation

Coordinator's configmap contains list of of oxia servers. To enable autoscaling for servers, coordinator needs to be updated with newly added or removed server pods. It's impossible w/o configurable config map name as suggested in this PR.

Also, fixed a bug about coordinator.tolerations and coordinator.nodeSelector.

### Modifications

### Verifying this change

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -520,7 +520,7 @@ oxia:
     # nodeSelector:
       # cloud.google.com/gke-nodepool: default-pool
     # customConfigMapName: ""
-    # entrypoint: {}
+    # entrypoint: []
 ## templates/server-statefulset.yaml
   server:
     # This is how Victoria Metrics or Prometheus discovers this component
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -118,7 +118,7 @@ components:
   # zookeeper
   zookeeper: true
   # oxia
-  oxia: true
+  oxia: false
   # bookkeeper
   bookkeeper: true
   # bookkeeper - autorecovery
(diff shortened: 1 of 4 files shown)


charts/pulsar/values.yaml
```diff
@@ -519,6 +519,8 @@ oxia:
     tolerations: []
     # nodeSelector:
       # cloud.google.com/gke-nodepool: default-pool
+    # customConfigMapName: ""
+    # entrypoint: {}
```
lhotari, line 523: should this be `entrypoint: []` instead?

lhotari (commented):
It seems that it would make sense to use the customConfigMapName in this template if it's set: https://github.com/apache/pulsar-helm-chart/blob/7833e51c282527e6b0c229f9bb5bae043e7a54f8/charts/pulsar/templates/_oxia.tpl#L103-L109 . In that case, there wouldn't be a need to set the entrypoint command. This would make it easier to use the customConfigMapName feature.

anosulchik:
> It seems that it would make sense to use the customConfigMapName in this template if it's set:
> 
> https://github.com/apache/pulsar-helm-chart/blob/7833e51c282527e6b0c229f9bb5bae043e7a54f8/charts/pulsar/templates/_oxia.tpl#L103-L109
> 
> . In that case, there wouldn't be a need to set the entrypoint command. This would make it easier to use the customConfigMapName feature.

@lhotari yeah, this is good idea - thanks, I've just submitted this change.

lhotari (approved):
LGTM

PR #607, opened 2025-05-05
add support of extra volumes and mounts for autorecovery

anosulchik:
Adds supports of extra volumes and volume mounts to bookkeeper autorecovery components.

### Motivation

autorecovery runs bookkeper so if the latter is configured with special volumes, it makes sense to support them in autorecovery component too.

### Modifications

### Verifying this change


lhotari (approved):
LGTM, thanks for contributing

PR #627, opened 2025-09-06
Correct Zookeeper URL to k8s

megaterrarc:
Fixes # 
 
Fixing URL for connecting to Zookeeper for k8s services

### Motivation

  Motivated by the helm chart installation outside the default namespace. Broker connection failure with zookeeper pulsar-tributaliza-zookeeper:2181, should be pulsar-tributaliza-zookeeper.namespace.svc.cluster.local:2181

### Modifications

 Changed the definition of define "pulsar.zookeeper.connect" to include the full service URL

### Verifying this change

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -1166,7 +1166,7 @@ broker:
       -XX:+ExitOnOutOfMemoryError
       -XX:+DisableExplicitGC
       -XX:+PerfDisableSharedMem
-    managedLedgerDefaultEnsembleSize: "2"
+    managedLedgerDefaultEnsembleSize: "1"
     managedLedgerDefaultWriteQuorum: "2"
     managedLedgerDefaultAckQuorum: "2"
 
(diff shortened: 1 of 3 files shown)


charts/pulsar/values.yaml
```diff
@@ -1166,7 +1166,7 @@ broker:
       -XX:+ExitOnOutOfMemoryError
       -XX:+DisableExplicitGC
       -XX:+PerfDisableSharedMem
-    managedLedgerDefaultEnsembleSize: "2"
```
lhotari, line 1169: This seems to be an unrelated change.

lhotari (commented):
> Motivated by the helm chart installation outside the default namespace. Broker connection failure with zookeeper pulsar-tributaliza-zookeeper:2181, should be pulsar-tributaliza-zookeeper.namespace.svc.cluster.local:2181

This should already be supported by the helm chart. It's very rare that the helm chart installation would happen in the default namespace. What is the scenario where you are facing the problem?

For example, the CI uses "pulsar" as the namespace.

Do you have some special DNS configuration (ndots) for your deployment?

PR #630, opened 2025-09-15
Add support for ClusterIssuer for certificates

### LINKED ISSUE #629: Allow use of ClusterIssuer for all component certs
In our environments we use a ClusterIssuer for all "in cluster" generated certs 
This allows clients to verify those certs as we supply a configMap of the CA cert (not key) in all namespaces

The issue with this helm chart is that it generates its own "CA" by requesting a cert/key from our ClusterIssuer and then uses that to generate an "Issuer" (namespace scoped) to then generate all component certs from.

So unless clients have access to the "CA" to insert into their chain they cannot verify the certs.

The fix would be to allow ALL component certs to use the configured ClusterIssuer 

This would mean it does not have to rotate its CA every 90 days and it cuts out extra components (the Issuer)

lhotari:
Fixes #629

### Motivation

Pulsar currently supports `kind: Issuer` issuers for certificates. There's also a need to support `kind: ClusterIssuer`.

### Modifications

- add `kind` and `group` keys to `certs.issuers.ca` key in `values.yaml`. 
- adapt the template to use these in the `issuerRef` for the certificate object

### DIFF
--- a/charts/pulsar/templates/tls-cert-internal-issuer.yaml
+++ b/charts/pulsar/templates/tls-cert-internal-issuer.yaml
@@ -55,6 +55,9 @@ spec:
     group: cert-manager.io
 ---
 {{- end }}
+{{- if and (eq .Values.certs.internal_issuer.type "ca") (not (eq .Values.certs.issuers.ca.kind "Issuer")) -}}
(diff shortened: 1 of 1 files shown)


lhotari:
> This still creates an Issuer that now is not required ?

I believe that changed with #565 so that it's not created unless `certs.internal_issuer.enabled` is set to `true`.

PR #654, opened 2026-02-24
Remove statsProviderClass config for bookkeeper

lhotari:
- prepare for 4.2.0 support since `statsProviderClass` has changed in Pulsar 4.2.0
- the default value for `statsProviderClass` is present in `conf/bookkeeper.conf` and there's no need to override it.


lhotari:
This is also required for 4.0.10 since we upgraded Jetty to 12.1.8 in 4.0.10 release.

PR #671, opened 2026-03-28
fix: Improve Zookeeper initialization wait logic to support multi url configuration store

### LINKED ISSUE #670: Cluster initialization fails when using multi url config store
**Describe the bug**
When using multi url configuration store, initialization doesn't happen because nslookup in wait-zk-cs-ready fails as it works only for a single URL

**To Reproduce**

- Deploy two instances of configuration store and configure as a single ensemble.
- Use that deployment as configuration store.
- Deploy helm chart.

**Expected behavior**
Cluster initialization should not fail as zookeeper is already initialized.

 - OS: MacOS

ganeshkalyank:
Fixes #670

### Motivation

When using a multi-URL configuration store (e.g., zk1:2181,zk2:2181), the wait-zk-cs-ready init container fails because nslookup cannot resolve comma-separated hostnames. This causes initialization to time out even when ZooKeeper is already accessible.

### Modifications

Replaced nslookup with bin/pulsar zookeeper-shell -server <configurationStore> ls /, which supports the full ZooKeeper connection string including multi-URL formats.

### Verifying this change


lhotari:
please check copilot's review comments

ganeshkalyank:
@lhotari addressed both the comments. also, assumed that configuration store deployment uses the same tls settings as zookeeper.

PR #672, opened 2026-04-13
Add support for generating JWT secrets during install

smbecker:
### Motivation

This eliminates the need for generating the JWT tokens outside of the helm chart before feeding the values into the chart.

### Modifications

Adds an optional pre-install hook that generates the necessary JWT tokens and saves them as the expected Kubernetes secrets

### Verifying this change


lhotari:
Thanks for the contribution @smbecker, very useful improvement.

charts/pulsar/templates/jwt-secret-init.yaml
```diff
@@ -0,0 +1,252 @@
...
+
+              create_secret() {
+                SECRET_NAME="$1"
+                shift
+                if output=$(kubectl create secret generic "$SECRET_NAME" -n "$NAMESPACE" "$@" 2>&1); then
+                  echo "Created secret $SECRET_NAME"
+                  return 0
+                elif echo "$output" | grep -q "already exists"; then
+                  echo "Secret $SECRET_NAME already exists, skipping"
+                  return 1
```
lhotari: together with `set -e`, this will currently cause the script to exit. Is that intentional?

charts/pulsar/templates/jwt-secret-init.yaml
```diff
@@ -0,0 +1,252 @@
...
+#
+
+# When JWT authentication is enabled and generateSecrets is true, this pre-install/pre-upgrade
+# hook job generates JWT signing keys and per-subject tokens, storing them as Kubernetes secrets.
+# The job is skipped entirely (not rendered) if the signing key secret already exists.
+# Individual token secrets that already exist are also left untouched.
+{{- if and .Values.auth.authentication.enabled .Values.auth.authentication.jwt.enabled .Values.auth.authentication.jwt.generateSecrets.enabled }}
+{{- $ns := include "pulsar.namespace" . }}
+{{- $keySecret := ternary "symmetric" "asymmetric" .Values.auth.authentication.jwt.usingSecretKey }}
+{{- if not (lookup "v1" "Secret" $ns (printf "%s-token-%s-key" .Release.Name $keySecret)) }}
```
lhotari, line 27: `lookup` returns empty during `helm template` and `helm install --dry-run`, so rendering them with the signing key  already present will still emit the Job. user who runs `helm template | kubectl apply` or
  any GitOps tool (f.e. ArgoCD) that doesn't have cluster context will rerun the generator and may overwrite their token files in `emptyDir` even when they don't want to. The current `lookup`-only guard isn't sufficient as the sole idempotency mechanism — at minimum, document that the feature requires actual `helm install/upgrade` against a live cluster, and consider letting `create_secret` succeed on conflict for the signing key too (or fail loud) instead of relying on render-time lookup.

charts/pulsar/templates/jwt-secret-init.yaml
```diff
@@ -0,0 +1,252 @@
...
+metadata:
+  name: "{{ template "pulsar.fullname" . }}-jwt-secret-init"
+  namespace: {{ $ns }}
+  labels:
+    {{- include "pulsar.standardLabels" . | nindent 4 }}
+    component: jwt-secret-init
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "0"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
```
lhotari, line 89: Job `hook-delete-policy` is `before-hook-creation,hook-succeeded` for the Job but `before-hook-creation,hook-succeeded,hook-failed` for the SA/Role/RoleBinding. Asymmetric: if the Job fails, the
  SA/Role/RoleBinding are deleted but the failed Job pod sticks around for debugging — and the next attempt will recreate the SA but  the Job remains, which on `before-hook-creation` will be replaced. It works, but it's worth either keeping `hook-failed` off the SA/Role/RoleBinding too (so the failed Job has the RBAC it needs for inspection) or adding it to the Job (so retries are clean).

lhotari (commented):
Please add an integration test to CI to validate the behavior. It might require multiple changes in `.ci` and `hack` directories besides `.github/workflows/pulsar-helm-chart-ci.yaml`. For example, you can add a new values file based on `.ci/clusters/values-jwt-asymmetric.yaml`. The CI scripts would require changes to disable the solution that creates the JWT tokens using the manual way. You can test the CI changes by enabling GitHub Actions in your own fork and creating a PR to your own fork to trigger the workflow run without waiting for maintainer approvals on apache/pulsar-helm-chart. In forks, it's also possible to ssh into the GitHub Action runner to investigate issues.

charts/pulsar/templates/jwt-secret-init.yaml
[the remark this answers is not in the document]
smbecker: This is intentional as regenerating this key on upgrades has caused issues for us the past.

charts/pulsar/templates/jwt-secret-init.yaml
```diff
@@ -0,0 +1,252 @@
...
+      nodeSelector:
+{{ toYaml .Values.auth.authentication.jwt.generateSecrets.nodeSelector | indent 8 }}
+    {{- end }}
+    {{- if .Values.auth.authentication.jwt.generateSecrets.tolerations }}
+      tolerations:
+{{ toYaml .Values.auth.authentication.jwt.generateSecrets.tolerations | indent 8 }}
+    {{- end }}
+      affinity:
+        {{- if and .Values.affinity.anti_affinity .Values.auth.authentication.jwt.generateSecrets.affinity.anti_affinity }}
+        podAntiAffinity:
```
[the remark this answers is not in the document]
smbecker: Moved outside of the `affinity` block

charts/pulsar/templates/jwt-secret-init.yaml
smbecker: No updated to `return 0`

charts/pulsar/templates/jwt-secret-init.yaml
smbecker: Correct that `lookup` will return empty without a cluster context. However, the existing `create_secret` behavior will effectively behave the same in that if the secret exists, it will skip over it. [That behavior](https://github.com/apache/pulsar-helm-chart/pull/672#discussion_r3113302156) is intentional as we have seen issues with regenerating this secret on upgrades.

charts/pulsar/templates/jwt-secret-init.yaml
smbecker: You can't do anything with the job besides review logs for debugging, which may be manually deleted by an operator before re-running. So, I set the delete policy on the RBAC resources to ensure that no matter what, those get cleaned up. I am happy to change it but that was the justification. Let me know if you still want the delete policy changed.

lhotari (approved):
LGTM, great job @smbecker

lhotari:
I already merged this PR, but would you also like to update README.md to reflect the changes made by this PR?
I guess `examples/values-jwt-symmetric.yaml` and `examples/values-jwt-asymmetric.yaml` could default to use the feature added by this PR.

smbecker:
> I already merged this PR, but would you also like to update README.md to reflect the changes made by this PR? I guess `examples/values-jwt-symmetric.yaml` and `examples/values-jwt-asymmetric.yaml` could default to use the feature added by this PR.

I don't see much that would need to change in the README, perhaps just updating the examples, which I can update if you want. The bigger change would need to happen in the [documentation site](https://pulsar.apache.org/docs/4.2.x/helm-deploy/#step-3-prepare-helm-release) where it directs you to use the `prepare_helm_release.sh` script to pre-generate the secrets.

PR #673, opened 2026-04-21
Add cert annotation configuration

smbecker:
### Motivation

This PR adds the ability to customize the annotations that are applied to the certificates that are generated by cert-manager. This can be useful for allowing tools like [reflector](https://github.com/emberstack/kubernetes-reflector) to propagate the certificate secrets to required namespaces.

### Modifications

Modifies the cert generation template to look for `annotations` property that gets merged into the `secretTemplate.annotations` property on the cert spec.

### Verifying this change


charts/pulsar/templates/_certs.tpl
```diff
@@ -75,6 +75,10 @@ metadata:
   labels:
     {{- include "pulsar.standardLabels" .root | nindent 4 }}
 spec:
+{{- if .tlsConfig.secretAnnotations }}
```
lhotari, line 78: Please document this by adding commented out lines in `charts/pulsar/values.yaml` for at least one of the "tlsConfigs" under the `tls:` key.

lhotari (commented):
Thanks for the contribution! One comment about documenting this new feature with a comment in `charts/pulsar/values.yaml`.

charts/pulsar/templates/_certs.tpl
lhotari: please also add a `secretAnnotations` config to `.ci/clusters/values-tls.yaml` to exercise this in CI. We don't have great validation for all features, so it's fine to just add the annotations to let's say `tls.proxy.secretAnnotations` in `.ci/clusters/values-tls.yaml` to validate that it doesn't error out after adding it.

smbecker: done

PR #675, opened 2026-04-24
Update standalone defaults to use zookeeper

smbecker:
### Motivation

Addresses comment from [previous PR](https://github.com/apache/pulsar-helm-chart/pull/674#issuecomment-4313610009)

### Modifications

Per the comment, default `PULSAR_STANDALONE_USE_ZOOKEEPER=1` when using standalone mode.

### Verifying this change

### DIFF
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -1225,6 +1225,7 @@ standalone:
       # storageClassName:
       selector: {}
   configData:
+    PULSAR_STANDALONE_USE_ZOOKEEPER: "1"
     PULSAR_MEM: >
       -Xms256m -Xmx512m -XX:MaxDirectMemorySize=512m
     PULSAR_GC: >
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -1225,7 +1225,6 @@ standalone:
       # storageClassName:
       selector: {}
   configData:
-    PULSAR_STANDALONE_USE_ZOOKEEPER: "1"
     PULSAR_MEM: >
       -Xms256m -Xmx512m -XX:MaxDirectMemorySize=512m
     PULSAR_GC: >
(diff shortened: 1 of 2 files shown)


charts/pulsar/values.yaml
```diff
@@ -1225,6 +1225,7 @@ standalone:
       # storageClassName:
       selector: {}
   configData:
+    PULSAR_STANDALONE_USE_ZOOKEEPER: "1"
```
lhotari: It might be better to put this in the configmap instead of the values.yaml file. In most cases there shouldn't be a need to use the RocksDB metadata store which might result in some incompatibilities since it's not extensively tested like Zookeeper is. There would be a comment in the configmap that setting the value to 0 will use RocksDB metadata store and the value shouldn't be changed after creating a cluster.

smbecker: > In most cases

To me, that sort of implies that there may be _some cases_ and putting it in the configmap prevents users from overwriting it. Instead, can I just add the necessary warning here?

lhotari: it should be possible to override the values that are present in the configmap

lhotari: https://github.com/apache/pulsar-helm-chart/blob/master/charts%2Fpulsar%2Ftemplates%2Fstandalone-configmap.yaml#L89-L91

smbecker: I may be misunderstanding what you are asking for. With where I put it in the values file, it would show up exactly where you linked. Can you clarify what you are asking for?

lhotari: It could be in the configmap template. as long as it's above the lines that I linked to, it can be overridden in values.yaml.

lhotari: I pushed the changes to this PR.

PR #680, opened 2026-04-29
Set FQDN to cluster domain instead of hard coded cluster.local

### LINKED ISSUE #679: Oxia does not pick up cluster domain configuration
**Describe the bug**
If the cluster is not configured to use cluster.local as the domain, bookkeepers get the wrong address to the Oxia instances.

**To Reproduce**
1. Create a new kubernetes cluster using cluster.test as the domain
2. Install pulsar

**Expected behavior**
Normal initialization and startup.

samkes:
Fixes #679

### Motivation
The cluster is not guaranteed to use cluster.local as its domain.

### Modifications
Remove the hard coded cluster.local value and use the same variable as everywhere else.

### Verifying this change


lhotari (approved):
LGTM

PR #695, opened 2026-06-19
Fix authentication examples, by enabling JWT.

### LINKED ISSUE #694: JWT examples missing required auth.authentication.jwt.enabled=true for token-keys mount
**Describe the bug**

The JWT example values should explicitly document/set `auth.authentication.jwt.enabled: true`.
Without it, the `proxy`, `broker`, etc. Pods does not mount `token-keys` (and related JWT secrets), which breaks JWT auth setup.

**Affected examples**

- `examples/values-jwt-asymmetric.yaml`
- `examples/values-jwt-symmetric.yaml`

**Expected**

JWT examples include (or clearly call out as required):

```yaml
auth:
  authentication:
    jwt:
      enabled: true
```

tallosim:
Fixes #694

### Motivation

`token-keys` is mounted only when `auth.authentication.jwt.enabled=true`.  
The JWT example values were missing this required flag, which can lead to broken JWT setups when users copy those examples.

### Modifications

Updated JWT example values to explicitly include the required setting:

- `examples/values-jwt-asymmetric.yaml`
- `examples/values-jwt-symmetric.yaml`

Added/kept `auth.authentication.jwt.enabled: true` so the examples match the chart behavior.

### Verifying this change

- [x] Render chart with JWT examples and confirm proxy Pod includes `token-keys` mount.


lhotari:
Thanks for the issue report and PR. I happened to address this while improving the examples.

PR #702, opened 2026-07-01
fix: omit empty broker clusterIP to avoid GitOps drift

### LINKED ISSUE #700: Broker Service clusterIP: "" causes perpetual Argo CD drift
## Problem

`broker-service.yaml` renders:

```yaml
clusterIP: "{{ .Values.broker.service.clusterIP }}"
```

The default value is `""`. After apply, Kubernetes assigns a real IP (e.g. `10.96.x.x`). Argo CD then always diffs desired vs live → perpetual OutOfSync

## Suggested fix

Omit `clusterIP` when the value is empty (same pattern as `standalone-service.yaml`):

```yaml
{{- if ne .Values.broker.service.clusterIP "" }}
  clusterIP: "{{ .Values.broker.service.clusterIP }}"
{{- end }}
```

yusufdaud-anaplan:
Omit clusterIP from the broker Service manifest when the value is empty, matching standalone-service.yaml. Prevents perpetual Argo CD OutOfSync when Kubernetes assigns a ClusterIP after apply.

Fixes apache/pulsar-helm-chart#700

### DIFF
--- a/charts/pulsar/templates/broker-service.yaml
+++ b/charts/pulsar/templates/broker-service.yaml
@@ -46,7 +46,7 @@ spec:
   - name: "{{ .Values.tlsPrefix }}pulsarssl"
     port: {{ .Values.broker.ports.pulsarssl }}
   {{- end }}
-  {{- if ne .Values.broker.service.clusterIP "" }}
+  {{- if .Values.broker.service.clusterIP }}
   clusterIP: "{{ .Values.broker.service.clusterIP }}"
   {{- end }}
   selector:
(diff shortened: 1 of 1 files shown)


charts/pulsar/templates/broker-service.yaml
```diff
@@ -46,7 +46,9 @@ spec:
   - name: "{{ .Values.tlsPrefix }}pulsarssl"
     port: {{ .Values.broker.ports.pulsarssl }}
   {{- end }}
+  {{- if ne .Values.broker.service.clusterIP "" }}
```
lhotari: If a user sets `broker.service.clusterIP: null` in their values (a common way to "unset" a chart default), Helm drops the key during value coalescing, so this expression becomes `ne nil ""`, which evaluates to true — rendering `clusterIP: ""` again and reintroducing the drift this PR fixes. A truthy check handles both `""` and `null`, while `"None"` and explicit IPs are non-empty strings and render as before:

```suggestion
  {{- if .Values.broker.service.clusterIP }}
```

lhotari (approved):

PR #709, opened 2026-07-14
Keep the internal CA private key across renewals with cert-manager 1.18+

### LINKED ISSUE #708: Self-signed CA gets a new private key on every renewal with cert-manager 1.18+, breaking TLS between components
**Describe the bug**

The CA Certificate created by the internal self-signed issuer (`tls-cert-internal-issuer.yaml`) doesn't set `spec.privateKey.rotationPolicy`. cert-manager 1.18 changed the default for that field from `Never` to `Always` (GA in 1.20, where it can no longer be turned off). With `Always`, every renewal of the CA certificate also generates a new CA private key.

The chart defaults the CA to `duration: 2160h` / `renewBefore: 360h`, so this happens roughly every 75 days. After a re-key, the component certificates are still the ones signed by the old key. As soon as a component reloads the new `ca.crt` (which works well since #613), it stops trusting the other components and handshakes fail with `PKIX path validation failed: Path does not chain with any of the trust anchors`. The cluster stays broken until cert-manager has re-issued every leaf certificate and every component has picked the new ones up.

On cert-manager older than 1.18 this never happened. Renewal reused the CA key, so certificates issued before the renewal kept validating.

**To Reproduce**

1. Install the chart with `tls.enabled=true` and `certs.internal_issuer.enabled=true` on cert-manager >= 1.18.
2. Force a CA renewal with `cmctl renew <release>-pulsar-ca -n <namespace>`, or wait for the renewal window.
3. Compare the public key in the CA secret before and after: it changes.
4. Watch broker/bookie/zookeeper handshakes fail until all leaf certificates have been re-issued.

**Expected behavior**

Renewing the CA certificate should not invalidate the certificates it issued. The CA Certificate should set `privateKey.rotationPolicy: Never`, which was the default when this template was written. A deliberate CA re-key is a separate concern and can already be handled with `tls.<component>.cacerts` (#619) by keeping the old and new CA in the truststore while certificates get re-issued.

dejanzele:
Fixes #708

### Motivation

Since cert-manager 1.18 the default `privateKey.rotationPolicy` is `Always` (GA in 1.20, can no longer be turned off). The chart's self-signed CA Certificate doesn't set the field, so the CA now gets a new private key on every renewal and all previously issued certificates stop validating until each of them has been re-issued. See #708.

### Modifications

- add `certs.internal_issuer.privateKey` to `values.yaml`, defaulting to `rotationPolicy: Never`
- render it as the `privateKey` block of the internal CA Certificate

`rotationPolicy: Never` keeps the CA key across renewals, which was the behavior before cert-manager 1.18. Setting `certs.internal_issuer.privateKey: null` renders the same manifest as before this change.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/tls-cert-internal-issuer.yaml
+++ b/charts/pulsar/templates/tls-cert-internal-issuer.yaml
@@ -45,6 +45,10 @@ spec:
     - server auth
     - client auth
   isCA: true
+{{- if .Values.certs.internal_issuer.privateKey }}
+  privateKey:
+    {{- toYaml .Values.certs.internal_issuer.privateKey | nindent 4 }}
+{{- end }}
   issuerRef:
     name: "{{ template "pulsar.fullname" . }}-{{ .Values.certs.internal_issuer.component }}"
     # We can reference ClusterIssuers by changing the kind here.
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -495,6 +495,13 @@ certs:
     duration: 2160h
     # 15d
     renewBefore: 360h
+    # Private key settings for the self-signed CA certificate.
+    # Keep rotationPolicy at Never so the CA key survives renewals. Since
+    # cert-manager 1.18 the default is Always, which generates a new CA key
+    # on every renewal and invalidates all certificates issued by the old key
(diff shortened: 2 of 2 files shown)


lhotari (approved):
LGTM

PR #714, opened 2026-07-30
Configure certificates subject

gulecroc:
### Motivation

Currently only certificates subject organizations is configurable.

This PR allow to configure full subject properties.

### Modifications

Move `tls.common.organization` to `tls.common.subject.organizations`.

Fail if  `tls.common.organization` is still used.

### Verifying this change

### DIFF
--- a/charts/pulsar/templates/_certs.tpl
+++ b/charts/pulsar/templates/_certs.tpl
@@ -87,9 +87,13 @@ spec:
 {{- end }}
   duration: "{{ .root.Values.tls.common.duration }}"
   renewBefore: "{{ .root.Values.tls.common.renewBefore }}"
+{{- if hasKey .root.Values.tls.common "organization" -}}
+{{- fail "tls.common.organization is no longer supported. Please configure tls.common.subject instead" -}}
+{{- end -}}
+{{- if .root.Values.tls.common.subject }}
   subject:
-    organizations:
-{{ toYaml .root.Values.tls.common.organization | indent 4 }}
+{{ toYaml .root.Values.tls.common.subject | indent 4 }}
+{{- end }}
   # The use of the common name field has been deprecated since 2000 and is
   # discouraged from being used.
   commonName: "{{ template "pulsar.fullname" .root }}-{{ .componentConfig.component }}"
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -268,8 +268,9 @@ tls:
     duration: 2160h
     # 15d
     renewBefore: 360h
-    organization:
-      - pulsar
+    subject:
+      organizations:
+        - pulsar
     keySize: 4096
     keyAlgorithm: RSA
     keyEncoding: PKCS8
--- a/charts/pulsar/values.yaml
+++ b/charts/pulsar/values.yaml
@@ -271,6 +271,13 @@ tls:
     subject:
       organizations:
         - pulsar
+      # countries: []
+      # organizationalUnits: []
+      # localities: []
+      # provinces: []
+      # streetAddresses: []
+      # postalCodes: []
+      # serialNumber: ""
     keySize: 4096
     keyAlgorithm: RSA
     keyEncoding: PKCS8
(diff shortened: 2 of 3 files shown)


charts/pulsar/templates/_certs.tpl
```diff
@@ -87,9 +87,13 @@ spec:
 {{- end }}
   duration: "{{ .root.Values.tls.common.duration }}"
   renewBefore: "{{ .root.Values.tls.common.renewBefore }}"
+{{- if hasKey .root.Values.tls.common "organization" -}}
+{{- fail "tls.common.organization is no longer supported. Please configure tls.common.subject instead" -}}
```
lhotari, line 91: I wanted to argue for a deprecation period here, then convinced myself the hard `fail` is right.

Without it, an existing `tls.common.organization: [acme]` would be silently dropped while the new default `subject.organizations: [pulsar]` survives Helm's merge — quietly changing the issued certificate's `O` with no signal. Failing loudly is the safer trade.

It's also consistent with how the chart already handles removed values (`auth.authentication.provider`, and `cert-manager.io/v1beta1` a few lines above this one).

On placement: I checked whether the guard being inside `pulsar.cert.template` makes it unreachable when no Certificate renders — it does (with `tls.enabled: false` the legacy key is silently ignored). But that's fine: in that state the value is genuinely inert, and enabling certificates later produces an immediate, actionable error. No change needed.

charts/pulsar/values.yaml
```diff
@@ -268,8 +268,9 @@ tls:
     duration: 2160h
     # 15d
     renewBefore: 360h
-    organization:
-      - pulsar
+    subject:
+      organizations:
```
lhotari, line 272: Could you list the other supported `X509Subject` fields here as comments? Right now `organizations` is the only one discoverable, so users won't know the rest exist:

```yaml
    subject:
      organizations:
        - pulsar
      # countries: []
      # organizationalUnits: []
      # localities: []
      # provinces: []
      # streetAddresses: []
      # postalCodes: []
      # serialNumber: ""
```

Worth doing because the block is a `toYaml` passthrough and the chart has no `values.schema.json` — so a typo like `subject.organization` (singular) gets pruned by the Certificate CRD's structural schema and issues a cert with no `O`.

lhotari (approved):
This looks good. I verified locally that the default render is byte-identical to master, the `subject` block maps correctly onto cert-manager's `X509Subject`, an empty `subject` renders nothing rather than `subject: null`, and `values.yaml` carries no `organization` remnant that could trip the guard spuriously.

Render recipe I used, if useful:

```
helm dependency build charts/pulsar
helm template test charts/pulsar --set "tls.enabled=true,tls.broker.enabled=true,\
tls.proxy.enabled=true,tls.zookeeper.enabled=true,tls.bookie.enabled=true,\
certs.internal_issuer.enabled=true,components.proxy=true" -s templates/tls-certs-internal.yaml
```

I initially wanted to push back on two things and talked myself out of both — see the inline notes.

### One request before merge: document the migration

The diff touches only `_certs.tpl` and `values.yaml`. The README's upgrade procedure explicitly says `helm get values … > values.yaml` and reuse it, which is exactly the path that now aborts. There's good precedent to follow: the "Upgrading to Helm chart version 4.1.0" section covers the structurally identical `auth.authentication.provider` removal.

Could you add a README "Upgrading to Helm chart version 4.8.0" section showing the before/after:

```yaml
# before
tls:
  common:
    organization:
      - pulsar
# after
tls:
  common:
    subject:
      organizations:
        - pulsar
```

Approving so this isn't blocked on me — please land the docs before merging.

### Note

#713 also edits `_certs.tpl`. `git merge-tree` shows no textual conflict at the current heads, but whichever lands second should be rebased and re-rendered.

<sub>Reviewed with Codex `gpt-5.6-sol` and Claude Opus 5; every finding reproduced locally by rendering the chart.</sub>

charts/pulsar/values.yaml
gulecroc: fixed in https://github.com/apache/pulsar-helm-chart/pull/714/commits/187ae7a24a6371f00fe4594ac92a5df88d38bd33

gulecroc:
Migration documented in https://github.com/apache/pulsar-helm-chart/pull/714/commits/79b949b8ca1a506cb627e59f91fb4a0b30ab79b8

